PR #2133, opened 2026-07-06
[salvage/product] Refactor desktop feature module registry

rodoHasArrived:
## Salvage provenance

Recreated stale product salvage work on top of current origin/main.

- Source branch: $sourceBranch
- Source commit: $sourceSha
- Original merge base: $base
- Current origin/main: $mainSha

## Validation

- git diff --cached --check passed before commit.
- Full ash scripts/ci.sh has not been run for this draft salvage PR.

## Changed paths

- src/Meridian.Wpf/Features/DesktopFeatureModuleRegistry.cs
- tests/Meridian.Wpf.Tests/Features/DesktopFeatureModuleRegistryTests.cs

### DIFF
--- a/src/Meridian.Wpf/Features/DesktopFeatureModuleRegistry.cs
+++ b/src/Meridian.Wpf/Features/DesktopFeatureModuleRegistry.cs
@@ -7,16 +7,16 @@ namespace Meridian.Wpf.Features;
 
 public static class DesktopFeatureModuleRegistry
 {
-    private static readonly IDesktopFeatureModule[] Modules =
+    private static readonly Func<IDesktopFeatureModule>[] ModuleFactories =
     [
-        new Home.HomeFeatureModule(),
-        new Trading.TradingFeatureModule(),
-        new Portfolio.PortfolioFeatureModule(),
-        new Accounting.AccountingFeatureModule(),
-        new Reporting.ReportingFeatureModule(),
-        new Strategy.StrategyFeatureModule(),
-        new Data.DataFeatureModule(),
-        new Settings.SettingsFeatureModule()
+        DesktopFeature.Module<Home.HomeFeatureModule>(),
+        DesktopFeature.Module<Trading.TradingFeatureModule>(),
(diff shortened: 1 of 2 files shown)


rodoHasArrived:
Closing — this refactors away a problem the code doesn't have, at a real allocation cost.

## The premise

The PR replaces the static array of module *instances* with an array of *factories*, so `GetModules()` constructs fresh modules on every call:

```csharp
public static IReadOnlyList<IDesktopFeatureModule> GetModules()
    => ModuleFactories.Select(static createModule => createModule()).ToArray();
```

That only buys determinism if the modules carry mutable instance state that could leak between callers. They don't:

- `IDesktopFeatureModule` (`src/Meridian.Wpf/Features/IDesktopFeatureModule.cs`) declares `Register`, `DescribePages`, `DescribeWorkspace`, and `DeclareCapabilities` — all pure reads apart from `Register`, which writes to the `IServiceCollection` passed in, not to the module.
- The implementations hold their page tables in `private static readonly IReadOnlyList<ShellPageDescriptor> Pages` (e.g. `Features/Home/HomeFeatureModule.cs:15`) — static and immutable.

Ordering determinism is already guaranteed twice over: the `Modules` array is a fixed literal, and `GetCapabilities()` closes with `.OrderBy(capability => capability.DisplayName, StringComparer.OrdinalIgnoreCase)`.

## The cost

`GetModules()` becomes O(8 allocations) per call instead of returning a cached array, and `GetCapabilities()` — which calls it — inherits that. `AddMeridianWpfFeatureModules` calls it during startup too. Small in absolute terms, but paid for no behavioural gain.

The `DesktopFeature.Module<T>()` helper also adds a `new()` constraint to every feature module, which forecloses constructor injection if a module ever needs it.

If there's a concrete state-leak or test-isolation failure behind this, please reopen with the failing case attached — that would change the analysis, since the fix shape here is reasonable for a module type that *does* hold state. As it stands the 47 lines of added tests assert a property the current code already has.

---

PR #2178, opened 2026-07-06
Consolidate shared UI primitives and shrink oversized screen roots

rodoHasArrived:
## Summary

Consolidate several duplicated/underused **shared UI primitives** in the browser workstation (`src/Meridian.Ui/dashboard`) and shrink two oversized screen roots. No behavior change except where noted (data-quality gains stale-response protection).

- **One `pluralizeCount()`** — added to `lib/format.ts` (with tests). The ~17 divergently-named local helpers (`formatCount`, `formatCountLabel`, `formatAccountingFocusCount`, `countUnit`, …) now delegate to it, so the `count === 1` branch lives in exactly one place. Names/signatures preserved → call sites and module-boundary guard tests untouched; output identical.
- **`useTableState` archived** — it had zero screen/view-model consumers (only a `concrete` barrel re-export + its own test). Removed with archive tombstones; its two still-used types (`TableSort`, `TableFilters`) moved to their sole consumer `lib/view-state-envelope.ts`. `AsyncRegion` was **kept** — it has an active consumer.
- **Governance keyed busy-map** — `accounting-screen.governance.view-model.ts`'s 23 separate `useState(false)` flags collapse into one keyed busy-map (`useGovernanceBusyMap`) with a typed key union and a single functional setter (concurrent ops no longer clobber). Thin derived aliases preserve every `<op>Busy`/`set<Op>Busy` reference and the returned shape.
- **Root-file shrink** — extracted downward-closed pure-helper clusters (no JSX, no back-references) into satellites: `reporting-screen.report-writer-helpers.ts` (reporting root 4021 → 3439) and `accounting-screen.approvals.ts` (accounting root 7531 → 7372).
- **`useRequestLifecycle` adopted** in `data-screen.data-quality.view-model.ts` — replaces a hand-rolled try/finally that had no stale-response guard; preserves the `{loading, error, model, refresh}` contract and adds renderHook tests (incl. a stale-discard test). First of the ~24 hand-rolled view-models; same pattern applies to the rest.
- **`ScreenLayout` doc reconciled** — it claimed to be "the one canonical skeleton every operator route inherits," but none of the 7 top-level workspace screens use it (they render bespoke `*WorkbenchContext` headers). Corrected the doc to describe it as the skeleton for secondary/detail routes, where all its consumers actually are.

## Reason

Remediation of a duplication/shared-primitive audit: a pluralization helper reimplemented under many names; `useRequestLifecycle` under-adopted; well-built primitives (`AsyncRegion`, `useTableState`) with no consumers; ~20 ad-hoc busy flags in one view-model; the documented canonical screen skeleton unused by top-level screens; and "split" screen files whose satellites supplemented rather than shrank the roots.

## Testing performed

- [ ] `bash scripts/ci.sh` completed successfully — **not run** (no local .NET toolchain in this environment; relying on GitHub `quality-gate`)
- [x] Relevant unit tests were added or updated (`pluralizeCount`, `useDataQualityPanel` renderHook tests)
- [ ] GitHub Actions `quality-gate` passed — pending
- [x] Dashboard `tsc --noEmit`, `eslint`, and the affected `vitest` suites pass locally (format, app-shell, reporting ×3, accounting-screen, governance VM, view-state-envelope, data-quality, + touched view-models — all green)

## Follow-up: server-side items from the same audit (need a .NET compiler/CI loop; not in this PR)

These were analyzed but intentionally **not** changed here — they touch security/auth and error-handling and cannot be compiled or unit-tested in this environment, so shipping them blind would be unsafe:

- **EndpointHelpers adoption / `ex.Message` leaks** — the genuine leak surface is **~36 unfiltered `catch (Exception ex)`** blocks that put `ex.Message` in the client response (top offenders: `ArchiveMaintenanceEndpoints` 17, `PackagingEndpoints` 7, `HistoricalEndpoints` 4). Most other `ex.Message` uses are **intentional** validation messages from typed/`when`-filtered catches and must be preserved. Note `FriendlyErrorFormatter` also surfaces raw `ex.Message`, so the fix is genericize-the-500-bucket, not just move code onto `EndpointHelpers`.
- **EndpointAuthorization** — `AccountingSystemEndpoints.cs` has 20 inline `if (!HasAccounting…) Forbidden()` guards (plus ~8 other files) that bypass the declarative `.RequirePermission()`/`.RequireAnyPermission()` helpers and are invisible to the route-level authorization-coverage check. Converting changes filter ordering (runs before model binding), so it needs the coverage test to validate.
- **`*ServiceBase` placement** — 11 of 13 are correctly-placed shared contracts for the deferred WPF shell (2 subclasses actually live in `Meridian.Ui.Services`). `AdminMaintenanceServiceBase` and `AdvancedAnalyticsServiceBase` have **no** subclasses and are used directly via DI/`.Instance` — the "Base" suffix is a misnomer; the fix is a cross-project rename (touches WPF DI + tests).

Happy to do these as a CI-validated follow-up.

## Safety review

## Governance changes

https://claude.ai/code/session_01GyPpDNq8hSmQ5W65JbaRRZ

---
_Generated by [Claude Code](https://claude.ai/code/session_01GyPpDNq8hSmQ5W65JbaRRZ)_


rodoHasArrived:
## CI status + review disposition

**CI failure is pre-existing and unrelated to this PR.** `verify-dotnet` / `verify-fast (restore/build/test)` / `quality-gate` are red on a single failing .NET test:

`Meridian.Tests.AssetOperations.PortfolioCashLadderEngineTests.Build_EarlyCallScenario_ReadsTermsBoundToSelectedRunNotNewestTerms` (`PortfolioCashLadderEngineTests.cs:193`)

The .NET **build succeeds with 0 errors / 0 warnings**; only that one assertion fails. This PR is **TypeScript-only** (`git diff --name-only origin/main..HEAD` contains no `.cs` files) and touches nothing in the portfolio cash-ladder engine — that test was last modified by `d8f3e842` on `main`, so the failure is inherited from the base branch, not introduced here. The browser/TS lanes (`verify-browser`, `verify-fast (browser workstation)`, Browser W4 Parity, CodeQL JS/TS) are green. No action taken on the ladder test as it's out of scope for this change.

**Gemini review comments:**

- ✅ **data-quality `catch` guard** (high) — applied in `018a542`. Added `if (!request.isCurrent()) return;` to the catch so a late-failing superseded request can't clobber newer state, symmetric with the success path (the underlying `safeSetState`/`fail` were already version-guarded, but the explicit early return is clearer). Locked with a new stale-failure test.
- ⏭️ **Optional-chaining suggestions** in `accounting-screen.approvals.ts` and `reporting-screen.report-writer-helpers.ts` — **not applied.** This code was moved **verbatim** from `accounting-screen.tsx` / `reporting-screen.tsx` as a behavior-preserving extraction (all 47 accounting + 106 reporting tests still pass). Adding null-guards would be a behavior change beyond the refactor's scope and would harden only the extracted copies while identical patterns remain at the call sites and elsewhere. If null-hardening against untrusted backend payloads is desired, it's better done as a separate deliberate pass at the source. (Note `dashboard/tsconfig.json` has `strict: false`, so these weren't type errors before or after the move.)

---

PR #2300, opened 2026-07-17
fix(docs-automation): repair broken command, stop CRLF churn, and end the API coverage feedback loop

rodoHasArrived:
Audit of the four docs-automation entrypoints:

```
python3 build/scripts/docs/run-docs-automation.py --profile core
python3 build/scripts/docs/generate-structure-docs.py --workflows-only
python3 build/scripts/docs/generate-workflow-manifest.py
node scripts/generate-diagrams.mjs
```

Two of the four were broken or actively harmful. This PR is **script fixes only** — no regenerated artifacts (see *Follow-up*).

> **Scope:** `build/scripts/docs/**` only, declared `phase:PR6` (the narrowest phase covering `build/**`). The `generate-diagrams.mjs` logging fix from this audit is split into its own PR — it sits outside the governed surface, so bundling it here would have forced a widen to `phase:PR9` for a six-line change.

## What was wrong

| Issue | Impact |
|---|---|
| `--workflows-only` required `--output` with no default | The documented command exited 2 — it never ran |
| `write_text()` uses the platform newline | Every Windows run rewrote LF-stored docs as CRLF, dirtying 5 tracked files with a **zero-content diff** |
| `"python3"` hardcoded for subprocesses | Locally `python3` is 3.9.13 but `python` is 3.11.5 — the orchestrator ran children under a *different* interpreter |
| `_git_visible_files` walked `.claude/worktrees/` | A git-ignored dir holding full repo checkouts injected **17,660 lines** into the generated tree |

## The API coverage dashboard measured its own output

`generate-api-contract-coverage-dashboard.py` scans `docs/**/*.md` for route paths to decide what is documented — and its own report lives in `docs/status/`, listing every endpoint it scans. An endpoint counted as documented purely because a **previous run wrote it there**.

That made the artifact non-convergent — a 2-cycle where each run's output became the next run's input:

| Branch | Committed | Corpus (= committed `.md`) | CI regen → | Observed diff |
|---|---|---|---|---|
| `main` | 100.0 (610 listed) | lacks `desktop-launch` | 99.8 | `-100.0 +99.8` |
| `claude/docs-regen-hygiene` | 99.8 (611 listed) | *has* it | 100.0 | `-99.8 +100.0` |

This is why **Documentation Automation has stayed red** even after that regeneration attempt, which got the failing diff down to exactly these two files and still failed.

Excluding generated report roots (`docs/status`, `docs/generated`) makes it reach a fixed point — three consecutive runs now produce byte-identical output.

### ⚠️ Reported coverage drops 100.0% → 29.3%

That is the true value (250/611 endpoints, 103/875 contracts). The self-reference was manufacturing **351 of the 610** "documented" endpoints — including `/api/auth/desktop-launch/{ticket}`, which no documentation mentions at all.

- Hand-written docs still count: `/api/system/shutdown` remains documented via `docs/development/process-lifecycle-diagnostics.md`.
- Nothing gates on the score; no other dashboard aggregates it.
- `generate-coverage.py` was checked for the same defect and converges cleanly.
- The ~361 now-visible undocumented endpoints are a real backlog this metric was hiding — not in scope here.

## Validation

- **99 tests pass, 3 new**, each verified non-vacuous (they fail with the fix neutralized).
- One pre-existing failure, `test_check_ai_inventory`, reproduces on a clean checkout of `main` and is unrelated to this branch.
- `--workflows-only` and the manifest generator both exit 0 and leave the tree clean, where previously one failed outright and the other dirtied five files.
- CI output is unchanged by the CRLF and worktree fixes (CI is Linux and has no worktrees); `docs-automation-summary.json` stays byte-stable because the recorded `python3` token is preserved.

## Expected check status

`regenerate-docs` will **fail on this PR**, as it already does on `main`. This PR does not turn that gate green — it removes the blocker that prevented it from ever going green. The dashboard fix legitimately enlarges that diff (committed `100.0` vs regenerated `29.3`), which the follow-up regeneration commit resolves.

## Follow-up (not in this PR)

1. **Regeneration commit** — now unblocked, but must run from a clean checkout.
2. `docs/generated/provider-registry.md` is **orphaned** — only `make/docs.mk --providers-only` builds it, not in any profile or CI, stale since 2026-06-06. Adopt into `core` or archive.
3. `enforceFreshness` on `docs-automation-core` uses file **mtime** — it can never fire in CI (fresh clones stamp checkout time) and false-positives locally on wall-clock. Suggest disabling; keep `enforceMissing`/`enforceCanonicalPaths`.

### DIFF
--- a/build/scripts/docs/generate-api-contract-coverage-dashboard.py
+++ b/build/scripts/docs/generate-api-contract-coverage-dashboard.py
@@ -44,8 +44,15 @@ DATA_SOURCES = [
     "src/**/*.cs endpoint mappings",
     "src/Meridian.Contracts/Api/UiApiRoutes.cs",
     "src/Meridian.Contracts/Workstation/*.cs",
-    "docs/**/*.md",
+    "docs/**/*.md (excluding generated report roots)",
 ]
+
+# Generated reports live under these roots and echo route paths and contract names verbatim —
+# including this dashboard's own markdown, which lists every endpoint it scans. Treating them as
+# documentation creates a feedback loop: an endpoint counts as documented purely because a prior
+# run wrote it into a report. That inflates coverage toward 100% and leaves the artifact unable to
+# converge, since each run's output changes the next run's input.
+GENERATED_DOC_ROOTS = ("docs/status", "docs/generated")
 
 
 def _should_skip(path: Path) -> bool:
@@ -161,6 +168,14 @@ def _scan_workstation_contracts(root: Path) -> list[dict[str, object]]:
     return contracts
 
 
+def _is_generated_doc(root: Path, path: Path) -> bool:
+    relative = _rel(root, path)
+    return any(
+        relative == generated_root or relative.startswith(generated_root + "/")
+        for generated_root in GENERATED_DOC_ROOTS
+    )
+
+
 def _load_docs_text(root: Path) -> str:
     docs_dir = root / "docs"
     if not docs_dir.is_dir():
@@ -168,6 +183,8 @@ def _load_docs_text(root: Path) -> str:
 
     chunks: list[str] = []
     for path in _iter_files(docs_dir, ".md"):
+        if _is_generated_doc(root, path):
+            continue
         chunks.append(_read_text(path))
     normalized = "\n".join(chunks)
     return ROUTE_CONSTRAINT_RE.sub(r"{\1}", normalized).casefold()
--- a/build/scripts/docs/generate-structure-docs.py
+++ b/build/scripts/docs/generate-structure-docs.py
@@ -44,6 +44,13 @@ EXCLUDED_ROOT_DIR_NAMES = {
     "data",
 }
 
+# Git-ignored directories that the filesystem merge in `_git_visible_files` would otherwise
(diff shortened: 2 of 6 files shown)


rodoHasArrived:
Closing as superseded — not abandoned.

The entire changeset from this PR landed on `main` via direct push `574c8b455` ("Harden docs automation and memory validation"). Verified byte-identical against current `main`:

- `build/scripts/docs/generate-api-contract-coverage-dashboard.py`
- `build/scripts/docs/generate-structure-docs.py`
- `build/scripts/docs/generate-workflow-manifest.py`
- `build/scripts/docs/run-docs-automation.py`

plus both new test files (`test_generate_api_contract_coverage_dashboard.py`, `test_generate_structure_docs.py`), which also show no diff against `main`.

`git diff origin/main origin/codex/docs-automation-hardening` now reports only *deletions* — this branch predates the SecurityMaster work that also arrived in `574c8b455`, so merging it would be a no-op at best and is misleading at worst.

No work is lost by closing. The docs-automation fix, the CRLF churn fix, and the API coverage feedback-loop fix are all live on `main`.

PR #2544, opened 2026-07-28
Activate the wash-sale engine and give realized gains a tax character

rodoHasArrived:
## Summary

Two connected gaps in tax-lot accounting.

**The wash-sale engine never ran.** `LedgerTaxLotReliefProjector.ComputeWashSale` is complete and well tested, but it returns `null` on every production disposal: `ComputeWashSale` requires an enabled policy *and* a non-empty replacement set, and every production construction of `LedgerTaxLotReliefInput` supplied neither — taking the `WashSalePolicy.Disabled` constructor default. `replacementAcquisitions:` appeared only in tests.

This adds the missing seam (`IWashSaleReplacementResolver`), durable policy, and durable deferral retention, so the engine actually fires.

**Realized gains carried no short/long-term character.** `Meridian.Backtesting.Sdk` knew about holding periods; the ledger — the side that feeds a 1099-B or K-1 — did not. Relief selections now carry `TaxCharacter`, holding-period days, and whether a prior wash sale extended the period.

One correction to the original diagnosis, worth flagging for review: the report pack's `DisallowedWashSaleLoss` column was not "always 0.00" — the artifact had **no rows at all**. `LedgerReportingAuthoritativeSource` called `LedgerReportPackBuilder.Build(ledger, reportRequest)` and passed no projections, so `tax-lot-realized-gains.csv` shipped as a bare header. Also, of the two call sites named as needing the resolver, only one is a real relief projection — `PostgresLedgerJournalStore.ValidateAuthoritativeDisposalPolicyAsync` uses the projector as a lot-selection *validator* with `salePrice: 0m` and reads only `relief.Selections`, so computing a wash sale there would be meaningless (every zero-price sale looks like a loss) and is deliberately not wired.

### Domain

- `WashSalePolicy` gains a replacement **scope** and an **effective date**. Scope defaults to `LedgerBook` — a ledger book is the accounting boundary for one taxable entity, which is the closest available analogue to the §1091 rule that looks across all of a taxpayer's accounts rather than one custodian. The effective date exists because enabling deferral changes reported losses; the projector now gates on `AppliesOn(saleDate)` rather than `Enabled`, so already-closed periods are not restated.
- Selections carry `TaxCharacter`, `HoldingPeriodDays`, and `HoldingPeriodExtendedByWashSale`, plus per-parcel `Proceeds` and `RealizedGainOrLoss` allocated so rows sum **exactly** to the projection totals rather than being re-derived (and drifting a cent) per consumer.
- Classification follows IRC §1222/§1223 — *more than* one year, compared calendar-wise. This deliberately differs from `ClosedLot.IsLongTerm`'s `>= 365 days`, which promotes a leap-year-spanning lot a day early; that rule was not propagated into tax reporting.
- `LedgerTaxLot` carries an optional holding-period start so a replacement inherits the disallowed sale's holding period under §1223(3). The carry chains: selling a lot that already absorbed a wash sale passes the older start along instead of resetting it.
- New `LedgerTaxLotBasisAdjustmentKind.WashSale` lets a retained deferral be replayed onto the replacement lot when it is finally relieved. It matches on lot identity rather than the acquired-before-effective-date guard the reference-data kinds use, because the §1091 window is symmetric and a replacement is usually *forward* of the sale.

### Storage

- Wash-sale policy is persisted on the existing effective-dated `tax_lot_policies` row, so relief method and deferral policy resolve together.
- New `wash_sale_deferrals` table retains each deferral against the replacement lot that absorbed it. It is the **single source of truth** — increased basis and carried holding period are not denormalized onto `tax_lots`, so they cannot drift out of step with a correction.
- `PostgresLedgerJournalStore` implements the resolver over a new `(ledger_book_id, security_id, acquired_date, lot_id)` index. The existing `ix_tax_lots_asset_scope_open` cannot serve this lookup: it is partial on open quantity (a fully-disposed lot is still a valid replacement) and orders `book_position_id` ahead of `acquired_date`, which a book-wide scope does not constrain.

### Reporting

- `LedgerTaxLotReliefHistoryProjector` rebuilds projections from retained disposal history. Proceeds are recovered as `basis + recognized + deferred` rather than stored separately, keeping rebuilt rows tied to the journal. The retained wash-sale outcome is reattached, not recomputed, so the pack reports what was booked rather than a re-derivation against today's lot history.
- A disposal whose retained economics cannot produce a well-formed projection is omitted rather than throwing — one bad historical row must not block a governed pack.
- The CSV adds `HoldingPeriodStartDate`, `HoldingPeriodDays`, `TaxCharacter`, `HoldingPeriodExtendedByWashSale`, and `ShortTermRecognizedGainOrLoss` / `LongTermRecognizedGainOrLoss`. The existing `RealizedGainOrLoss,DisallowedWashSaleLoss,RecognizedGainOrLoss` header substring is preserved contiguously.

### Bug fixed along the way

`SecurityMasterCostBasisAdjustmentService.ProjectReliefAsync` rebuilt `LedgerTaxLotReliefInput` without `washSalePolicy` or `replacementAcquisitions`, so a caller that had resolved replacements lost them the moment it routed through Security Master enrichment and silently recognized a loss the policy meant to defer.

## Reason

The wash-sale work was already built and tested but structurally unreachable in production, and the governed report pack advertised a wash-sale column it could never populate. For a platform whose pitch is "prove the number," a confidently-formatted empty artifact is the expensive kind of gap.

## Testing performed

No .NET SDK is available in this environment, so nothing was compiled or run locally — validation is entirely via the hosted workflows, and the boxes above stay unchecked until those pass. Please treat the build result on this PR as the first real signal.

New unit tests cover the holding-period boundary (including a leap-year span a 365-day count would promote a day early), character totals partitioning the realized total exactly, parcel proceeds summing to the projection total, the dated policy gate in both directions, carry-date chaining through an already-extended lot, and history rebuild with and without a deferral.

Not covered by tests here: the Postgres resolver and disposal-history queries have no integration coverage in this change — they need a database fixture. The pure reconstruction logic they feed is tested directly.

## Safety review

## Governance changes

---
_Generated by [Claude Code](https://claude.ai/code/session_01UTZH5MbMNJJ79KvMbPcg6T)_

### DIFF
--- a/src/Meridian.Application/SecurityMaster/SecurityMasterCostBasisAdjustmentService.cs
+++ b/src/Meridian.Application/SecurityMaster/SecurityMasterCostBasisAdjustmentService.cs
@@ -100,6 +100,10 @@ public sealed class SecurityMasterCostBasisAdjustmentService : ISecurityMasterCo
         // ahead of the ones sourced from the Security Master.
         var combined = baseInput.BasisAdjustments.Concat(adjustments).ToArray();
 
+        // Every field of the base input has to be carried forward, not just the adjustments: the
+        // wash-sale policy and replacement acquisitions were previously dropped here, so a caller
+        // that resolved replacements lost them the moment it routed through Security Master
+        // enrichment and silently recognized a loss the policy meant to defer.
         var enrichedInput = new LedgerTaxLotReliefInput(
             baseInput.Account,
             baseInput.SaleDate,
@@ -109,7 +113,9 @@ public sealed class SecurityMasterCostBasisAdjustmentService : ISecurityMasterCo
             baseInput.OpenLots,
             baseInput.FinancialAccountId,
             baseInput.SpecificLotIds,
-            combined);
+            combined,
+            baseInput.WashSalePolicy,
+            baseInput.ReplacementAcquisitions);
 
         return LedgerTaxLotReliefProjector.Project(enrichedInput);
     }
--- a/src/Meridian.Ledger/IWashSaleReplacementResolver.cs
+++ b/src/Meridian.Ledger/IWashSaleReplacementResolver.cs
@@ -1,0 +1,69 @@
+namespace Meridian.Ledger;
+
+/// <summary>
+/// The question a wash-sale replacement search answers: for this loss-generating disposal, which
+/// acquisitions of the same security fall inside the policy window, and which prior deferrals have
+/// already been capitalized into the lots being relieved?
+/// </summary>
+/// <param name="LedgerBookId">The disposing ledger book; also the search boundary for <see cref="WashSaleReplacementScope.LedgerBook"/>.</param>
+/// <param name="DisposingAccount">The account the disposal relieves lots from.</param>
+/// <param name="SecurityId">Security Master identity of the sold security. Replacements must be substantially identical.</param>
+/// <param name="SaleDate">Sale date the ±<see cref="WashSalePolicy.WindowDays"/> window centres on.</param>
+/// <param name="Policy">The policy in force for this sale; supplies the window and the account scope.</param>
+/// <param name="RelievedLotIds">
+/// Lots being relieved by this disposal. They are excluded from the replacement candidates — a lot
+/// cannot be its own replacement, and matching one would disallow a loss against itself.
+/// </param>
+public sealed record WashSaleReplacementQuery(
+    Guid LedgerBookId,
+    LedgerAccount DisposingAccount,
+    Guid SecurityId,
+    DateOnly SaleDate,
+    WashSalePolicy Policy,
+    IReadOnlyList<string> RelievedLotIds);
+
+/// <summary>
+/// What a <see cref="IWashSaleReplacementResolver"/> found for one disposal.
+/// </summary>
+/// <param name="Replacements">
+/// Candidate replacement acquisitions inside the policy window. The relief engine still filters
+/// these by window and security, so a resolver may return a slightly wider set without changing the
+/// computed deferral.
+/// </param>
+/// <param name="PriorDeferrals">
+/// Wash-sale deferrals already capitalized into the lots this disposal is relieving, expressed as
+/// <see cref="LedgerTaxLotBasisAdjustmentKind.WashSale"/> basis adjustments. Feeding these back in
+/// is what makes a deferral survive to the sale that finally recognizes it: the replacement lot is
+/// relieved at its increased basis and with the carried holding period, rather than at the raw
+/// price it was bought for.
+/// </param>
+public sealed record WashSaleReplacementLookup(
+    IReadOnlyList<WashSaleReplacementAcquisition> Replacements,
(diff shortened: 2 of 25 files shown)


rodoHasArrived:
## `schema-control` is red on the base branch, not on this diff

I investigated this rather than assuming, and I'm not going to "fix" it here — committing what I'd have to commit would make things worse. Evidence:

**It fails without my change.** I ran the verifier against unmodified `origin/main` in a worktree with a clean PostgreSQL 16 database:

```
python build/scripts/schema-control.py --root <main-worktree> verify --database-url ... --base-ref origin/main
→ Committed schema-control manifests or documentation are stale.
```

**CI agrees.** Recent `schema-control` runs:

| Result | Date | Branch |
| --- | --- | --- |
| failure | 2026-07-28 | `claude/wash-sale-engine-activation-fejutr` (this PR) |
| failure | 2026-07-28 | `claude/portfolio-aware-risk-engine-f0hrrq` (×11 today) |
| failure | 2026-07-28 | `dependabot/github_actions/actions/checkout-7.0` |
| failure | 2026-07-27 | **`main`** |

The dependabot run is the clincher — an `actions/checkout` version bump cannot touch a migration. The last green runs were 2026-07-25/26.

**Why it went unnoticed:** the workflow triggers on `pull_request` filtered to migration/contract paths, so base-branch staleness stays invisible until a PR touches those paths. This PR adds `V_ledger_028__wash_sale_activation.sql`, so it's the messenger.

**Why I'm not committing a regenerated snapshot.** I did regenerate it locally to check. Two problems:

1. **Most of the diff isn't mine.** A snapshot of *unmodified main* in this environment produces 34,488 insertions / 14,585 deletions on its own — including ~19k lines of `reporting_access_grants` constraint-dependency churn. Isolating my migration's actual delta (diffing a main-snapshot against a branch-snapshot from the same toolchain) gives ~4.1k added lines touching only `catalog/dependencies/migrations/policies`, `schemas/ledger.json`, and the ledger diagram/module docs — no `reporting` churn at all. So the reporting churn is baseline drift, not a consequence of this change.
2. **I can't produce artifacts valid for CI.** This environment has PostgreSQL 16.13; CI uses 16.14. The manifests carry content fingerprints over the whole catalog, so anything I generate here would be fingerprinted for the wrong server and would still fail `verify`. Patching my isolated delta onto the committed baseline also fails — the context lines and hashes in `schema-catalog.md` don't match, precisely because the committed baseline is stale.

I've reverted all of it; the branch carries no manifest changes.

**Suggested fix, outside this PR:** run the workflow's own refresh path against a clean base branch —

```
gh workflow run schema-control.yml --ref main -f mode=snapshot
```

— and commit the result. Once `main` is green, this PR's migration delta should regenerate cleanly on top. Happy to do that in a separate PR if you'd like; it just shouldn't ride along here, where it would bury a ~4k-line real change under ~35k lines of unrelated churn.

The lane that actually gates this change is `Meridian CI / quality-gate`; I'm tracking that separately.

---

PR #2548, opened 2026-07-31
Adopt the 2026-07 depth slate as W10 roadmap rows

rodoHasArrived:
## Summary

Adds eleven `W10` rows to the roadmap registry as an accepted depth slate, plus the decision-log, risk-register, document-index, and program-state entries that bind them, and a slate rationale doc. No product code changes — registry and planning only.

The slate came out of a brainstorm over current source. Its organizing finding is that the recurring gap is **not missing capability — it is capability that was built, tested, and never connected.** Six of the eleven rows wire existing code rather than write new code:

- **Bulk break resolution exists end to end and is completely unwired.** `ApplyBulkCaseworkAsync` (idempotency key, bounded case count, dry-run, partial-success, retained receipts), the mapped endpoint, and three browser client functions all exist. `ReconciliationBreakBulkAction` has zero hits across `src/Meridian.Ui/dashboard/src/` outside `lib/api.ts`.
- **The Number Passport component never fetches provenance.** It infers all eight rows by keyword-scanning relationship labels. `LedgerAmountProvenanceService` has no client consumer in the browser workstation, `Meridian.Ui.Services`, or `Meridian.Wpf`.
- **`FundAdministrationControlService` is not registered in DI anywhere** — only tests reference it — and holds its schedules, template book, locked-period book, and posted-occurrence guard in memory.
- **The wash-sale/tax-character engine terminates in one report-pack CSV** that no endpoint serves and no screen reads; `TaxCharacter` has zero references in `Meridian.Contracts`, `Meridian.Ui`, or `Meridian.Wpf`.

| # | Row | Kind |
| -- | --- | --- |
| 1 | `W10-MARK-001` — Fail-closed stale-mark policy and mark-age surfacing | activation |
| 2 | `W10-RECON-001` — Durable break lineage identity and run-over-run diff | new seam |
| 3 | `W10-PROV-001` — Ledger-amount evidence subject and shared proof drawer | activation |
| 4 | `W10-RECON-002` — Break clustering and bulk-resolution activation | activation |
| 5 | `W10-JRNL-001` — Durable recurring journal schedules and draft runner | activation |
| 6 | `W10-TAX-001` — Tax character, wash-sale, and lot-relief operator surface | activation |
| 7 | `W10-SEAM-001` — Unified close-readiness projection behind one shared contract | consolidation |
| 8 | `W10-RECON-003` — Unified tolerance model and what-if replay workbench | new capability |
| 9 | `W10-RECON-004` — Operator-taught match rules with promotion gate | new capability |
| 10 | `W10-PERF-001` — Portfolio and investor return measurement | new capability |
| 11 | `W10-CONSOL-001` — Intercompany elimination on consolidated ledger views | new capability |

Each row enters as `planned` with `planned_evidence` posture, so nothing here reads as a completion claim. `production_readiness.status` remains `blocked`; `W10-MARK-001` and `W10-SEAM-001` are pulled forward only because they serve the release gate.

## Reason

Requested as a sequenced implementation plan across the full brainstorm output. The registry is the durable home for that sequencing — a plan living only in a chat thread cannot gate work or carry dependencies.

## Review

Eleven review rounds from `chatgpt-codex-connector` and `copilot-pull-request-reviewer` produced **64 findings. Every one was valid; none were rejected.** Round-by-round detail is in the thread. Three themes are worth recording here:

**1. Rows stated design, and design attracts design review.** The rows initially carried 7–9 mechanism-shaped criteria naming keys, guards, and storage shapes. They were trimmed to 5–6 outcome-shaped criteria matching W9 house style, with every verified source fact moved to a **Known Constraints from Source** section in the slate doc, attributed to the type that owns the behavior. That section is now the most useful artifact in this PR and the place an implementer should start.

**2. Five findings were violations of this repo's own design charter, not of code.** Nine rounds of verifying claims against owning types did not surface them, because the charter governs what the code is *allowed* to do:

| Row | Charter rule |
| --- | --- |
| `W10-RECON-004` | §21 straight-through lanes require versioned policy, materiality caps, reversible evidence, sampling review, kill switch, per-item material handling |
| `W10-RECON-003` | same — committing a tolerance auto-suppresses future breaks, so it is a policy change, not a preference |
| `W10-MARK-001` | §11.5 override strategy requires prices to be reviewed or expired, not merely dated and approved |
| `W10-CONSOL-001` | concurrency control requires stale-version fail-closed; posted immutability requires linked reversing or adjusting journals |

**3. A recurring shape: fixing the mechanical defect and leaving the governance one.** Twice — `W10-RECON-004`, then `W10-RECON-003` — a preview or promotion gate was bound correctly against drift while conferring authority nobody had approved. Both now carry the charter conditions.

## Ordering in generated views

W10 numbers per area, so every identifier ends in `001` and the trailing number carries no rank. The renderer's tie therefore fell through to an alphabetical area sort, and because it chains items with `-->`, the diagram drew rank-11 `CONSOL` as a prerequisite for rank-5 `JRNL`. Both reviewers raised it independently.

Fixed in **#2551** (merged), which added an optional `sequence` field and gave the diagram, `ROADMAP_SUMMARY.md`, and `roadmap-register.md` one shared wave-then-rank ordering. The W10 rows here now declare `sequence: 1..11`, and all three views render the adopted order:

```
MARK(1) → RECON-001(2) → PROV(3) → RECON-002(4) → JRNL(5) → TAX(6)
→ SEAM(7) → RECON-003(8) → RECON-004(9) → PERF(10) → CONSOL(11)
```

The schema header advances to `meridian.roadmap-items@1.1.0` / `minimum_renderer_version: 2.0.0`, both inherited from #2551 through the merge from `main`.

## Testing performed

- [x] `bash scripts/ci.sh --lane verify-docs` — exit 0, 0 drift, against the committed tree
- [ ] Relevant unit tests were added or updated — n/a, registry data change
- [ ] Relevant integration tests were added or updated — n/a, registry data change
- [ ] GitHub Actions `quality-gate` passed — pending on this head

```
python3 build/scripts/docs/validate-roadmap-registry.py --summary   # 0 errors, 0 warnings
python3 build/scripts/docs/render-roadmap-docs.py --summary
python3 build/scripts/docs/render-roadmap-diagrams.py --summary
python3 build/scripts/docs/run-docs-automation.py --profile core ...
python3 scripts/check_program_state_consistency.py                  # passed
```

39 rows (28 + 11), all eleven `W10-*` `planned` / `green`, 5–6 exit criteria each, snapshot `2026-07-31`.

**On earlier `verify-dotnet` failures:** two were the month-end schedule bug fixed in #2549; two were cancellations from pushing into a live run; two were pre-existing async races in `MaintenanceSchedulerTests` and `OrderManagementSystemReportStreamTests`, confirmed flaky when an identical tree passed on the next run. Those two races are real and unfixed — they belong on their own branch, not in a `phase:PR1` registry PR that cannot touch `tests/**`.

## Safety review

## Governance changes

Registry data, product docs, and regenerated outputs only.

## Open questions for the owner

1. **`W10-RECON-003` is the weakest row.** Its two prerequisites — unifying three incompatible tolerance shapes (the engine consumes only the flattest, so price, basis-point, and settlement-date rules are structurally unreachable) and retaining replayable run artifacts including the internal side — are each larger than the headline preview. Splitting or deferring it from v1 is a live option.
2. **`W10-SEAM-001` needs a sequencing handshake with `W8-WPF-PARITY-001`** before its phase gate closes, or the desktop lane duplicates the client-side readiness aggregation instead of retiring it.
3. **`W10-MARK-001` changes behavior rather than preserving it**, and the affected policy type is public, so the default flip is breaking for external constructors. It carries a preview requirement; the flip itself is an owner decision.


rodoHasArrived:
## CI status after `4e7a6deb6`

Two failures on the first run. One was mine and is fixed; the other is red on the base branch and is not caused by this PR.

### `regenerate-docs` — fixed in `4e7a6deb6`

Mine. `docs/status/example-validation.md` is generated by `build/scripts/docs/validate-examples.py`, and adding the slate doc's one `bash` block changed its scan counts (361 → 362 files, 958 → 959 blocks). I regenerated it via the same profile CI uses:

```
```

That run also picked up **pre-existing drift already on `main`**: `repository-structure.md` and `coverage-report.md` were missing the wash-sale activation sources (`TaxCharacter.cs`, `IWashSaleReplacementResolver.cs`, `LedgerTaxLotReliefHistoryProjector.cs`, the `PostgresLedgerJournalStore` wash-sale partials, migration `V_ledger_028`). Those files exist on `main` but were never rendered into the generated docs:

```
$ git cat-file -e origin/main:src/Meridian.Ledger/TaxCharacter.cs && echo present
present
$ git show origin/main:docs/generated/repository-structure.md | grep -c TaxCharacter.cs
0
```

So that extra churn in the diff is a catch-up, not a change of mine. Flagging it because it makes the second commit look wider than the slate itself. `bash scripts/ci.sh --lane verify-docs` now exits 0 locally against the committed tree.

### `verify-dotnet` — pre-existing on `main`, not fixing here

Two tests fail in `core-application`:

```
Meridian.Tests.Application.SecurityMasterCashFlowServiceTests
  .GetProjectionAsync_CalculatedSinker_ShouldAmortizePrincipalAcrossScheduleDates
  .GetProjectionAsync_SingleFloatingLeg_ProjectsFloatingRateNoteWithPrincipalExchange

Expected projection.Schedule to contain 4 item(s), but found 5
  first entry PeriodDate = <2026-10-31 +0h>
```

This PR changes **only files under `docs/`** — `git diff --name-only origin/main...HEAD | grep -v '^docs/'` is empty — so it cannot affect a cash-flow projection.

The tests seed their as-of date from wall-clock UTC:

```csharp
// tests/Meridian.Tests/Application/SecurityMasterCashFlowServiceTests.cs:63
var asOf = DateOnly.FromDateTime(DateTime.UtcNow.Date);
```

Today is 2026-07-31, a month end, which shifts a period boundary and yields one extra schedule entry. The scheduled runs on `main` confirm it — **the same commit `2ed06af6` that this branch is based on passed on 2026-07-28 and has failed every nightly since 2026-07-29**:

| Run date | SHA | Conclusion |
| --- | --- | --- |
| 2026-07-31 | `2ed06af6` | failure |
| 2026-07-30 | `2ed06af6` | failure |
| 2026-07-29 | `2ed06af6` | failure |
| 2026-07-28 | `2ed06af6` | success |

Identical code, different day. `quality-gate` is failing downstream of this.

I'm not fixing it in this PR: it is unrelated to the slate, and a date-dependent projection test needs an owner who can decide whether the expected schedule or the boundary handling is wrong — that is a behavior question, not a test-data tweak. Worth noting it is exactly the failure mode `W10-RECON-003`'s exit criteria call out (a determinism audit before any wall-clock-dependent result is shown), and the kind of thing a test-isolation pass would catch.

Happy to open a separate issue or fix it on its own branch if you'd like.

---

rodoHasArrived:
The `verify-dotnet` / `quality-gate` failures on `4e7a6deb6` are the pre-existing month-end schedule bug described above, not this PR. Fix now open as #2549 — it anchors payment dates on the issue date instead of compounding `AddMonths` from the previous (clamped) payment date.

This PR stays red on `verify-dotnet` until #2549 lands on `main`; everything it actually touches (`regenerate-docs`, `validate-docs`, `verify-docs`, `verify-workflows`, `schema`, `scope-gate`) is green.

---

rodoHasArrived:
Reviewed all six findings against the source. **All six are valid.** Five are fixed in `9b17f95be`; one is deliberately deferred with a reason.

### The one that corrects me

`FundAccountEndpoints.cs:471` maps `GET /{accountId:guid}/performance`, backed by `BrokeragePortfolioSyncService.GetPerformanceAsync`, returning `BrokeragePortfolioPerformanceDto`. It has **no constant in `UiApiRoutes.cs`** — registered inline — which is exactly why my route-constant search missed it. The "no performance route" premise was wrong and I've corrected it.

Reading the implementation, the existing measure is `endingEquity - beginningEquity - netCashFlow` over brokerage balance snapshots, with the percentage divided by beginning equity — a **single-period simple return for one linked account**. So the remaining gap is narrower and more precise than I stated: no ledger-derived return, no time-weighted return with sub-period chaining, no money-weighted return honoring cash-flow timing, no investor-level return over capital-account activity. `W10-PERF-001` now says exactly that, and carries a new first exit criterion requiring the brokerage seam to be **extended or explicitly superseded with the reason recorded**, so nobody builds a parallel API by accident. Your suggested framing was right.

### Also fixed

- **Lineage key in the cluster key** — correct, and it was self-defeating: `W10-RECON-001` defines lineage as the identity of one break, so every cluster would have held exactly one member. Grouping is now `(break type, counterparty, currency, sign, variance band)`, with a new criterion stating lineage is retained *on members* for diffing and evidence.
- **Promotion on raw hit counts** — correct. Promotion now requires operator-adjudicated shadow outcomes, a minimum sample size, and a precision or false-positive bound, with an explicit criterion that a raw hit count never substitutes for an adjudicated outcome.
- **PRD mappings** — correct on all three. `PRD-000` is deployment-posture parsing and startup policy, `PRD-010` is ingress security and bounded parsing, `PRD-017` is documentation automation and same-commit drift. I **removed** the three relationships rather than inventing replacements, and noted why in the table — an empty cell is honest; a wrong mapping lets release planning count W10 work against unrelated blockers.
- **`registry.snapshot_date`** — advanced to `2026-07-31`. Generated summary now reads `Snapshot date: 2026-07-31`.
- **Orphaned slate doc** — linked from `docs/product/README.md` beside the W9 slate entry. Orphan count is back to **258**. I'd seen that 258→259 move in the diff and didn't act on it; good catch.

### Deferred, with reason

**The diagram ordering.** Confirmed, and worse than described — the generated `.mmd` is a single lexicographic chain, so it contains `W1_DATA_001 --> W10_CONSOL_001` (line 4), opens W10 with rank-11 `CONSOL`, and has `W10_TAX_001 --> W2_PROMO_001` (line 26): a planned row pointing into completed W2 work.

The fix belongs in `render-roadmap-diagrams.py` — natural-sort wave tokens so `W1 < W2 < … < W9 < W10` (with `W5X` after `W5`) and order within a wave by slate rank. **That changes how every wave renders, not just W10**, which makes it a generator behavior change inside a PR whose stated scope is registry data and whose safety review claims no unrelated changes. Splitting it out is the same call I made for the cash-flow bug in #2549, and it keeps this PR reviewable.

Flagging it here rather than leaving it silent: until that lands, the generated roadmap diagram asserts a delivery sequence that contradicts the adopted slate.

Validation after the corrections: `validate-roadmap-registry` 0 errors, `check_program_state_consistency` passed, `scripts/ci.sh --lane verify-docs` exit 0, 39 rows still rendering with all eleven `W10-*` `planned` / `green`.

---

rodoHasArrived:
The deferred diagram-ordering finding is now open as #2550, off `main`.

Fixing it turned up a second defect that predates this PR: items inside a wave sorted alphabetically by area, so W9 rendered as `ALPACA, ASSET, DEMO, GOV, …` rather than in the rank its numeric suffixes encode per `DEC-PRIORITY-SLATE-001`. #2550 fixes both — waves sort numerically (`W5 < W5X`, `W9 < W10`) and items within a wave sort by sequence — with three regression tests.

One limitation worth knowing here: #2550 cannot render the W10 slate in *its* adopted rank, because W10 identifiers deliberately don't encode rank the way W9's do — that was the point of opening a new wave token rather than extending W9. W10 will render deterministically but alphabetically within the wave. Carrying slate rank into the diagram would need an optional ordering field on the registry schema, which I've left as a follow-up rather than folding into a rendering fix.

---

rodoHasArrived:
Round two: all three findings verified and valid. Fixed in `d82dc2536`.

**Stale marks vs. the simulation risk.** You're right, and the distinction matters for burn-down: an aged mark is *real data presenting as current*, not *simulated data presenting as real*. Attaching `W10-MARK-001` to `RISK-SIM-REAL-001` would have let stale-mark work read as simulation-safety progress. Took the "separate risk" option rather than dropping the relationship — added `RISK-STALE-MARK-001` ("Aged valuation inputs can price current outputs without disclosure") and repointed the row there, plus the two prose spots that repeated the old claim.

**`SRC-BACKTESTING` on `W10-PERF-001`.** Correct — the exit criteria require lifting `XirrCalculator` out of `src/Meridian.Backtesting/Metrics/`, so the module that has to change was missing from traceability while only the destination was listed. Added, keeping `SRC-FSHARP` as the destination.

**`W10-CONSOL-001` describing treatment kinds as journal sources.** Confirmed against `LedgerBookDtos.cs`: `Intercompany = 12` and `ConsolidationElimination = 13` are `AccountingTreatmentKindDto` members, reached through `AccountingPolicyRuleDto.TreatmentKind`.

Reading that record made the correction stronger than a wording fix. `AccountingPolicyRuleDto` already carries `JournalTemplateId`, `RequiresEvidence`, `RequiresApproval`, and `AllowsAutoPosting` — which is precisely the seam for producing elimination drafts under approval, and exactly what the row wanted. So the current-state premise is recast around the policy rule, and there's a new first exit criterion requiring the treatment kind to be driven by a rule on that seam **with no parallel source or treatment discriminator introduced beside it**.

Validation: `validate-roadmap-registry` 0 errors, `check_program_state_consistency` passed, `scripts/ci.sh --lane verify-docs` exit 0.

Separately, the diagram-ordering finding from round one is now open as #2550.

---

rodoHasArrived:
Round three: all six verified against source and valid. Fixed in `4a7b82346`, plus `main` merged in so `verify-dotnet` picks up #2549.

### The P1 that changes what the row can promise

**Mixed-lot wash-sale math.** This is the most important finding of the review, and it isn't a wording problem. `LedgerTaxLotReliefProjector.ComputeWashSale` guards on `realizedGainOrLoss >= 0m`, and its own XML doc says it outright:

> *"A single relief that spans both gain and loss lots (a net gain that nonetheless contains loss shares) is not yet decomposed to defer only the loss shares; per-loss-lot matching within a mixed gain/loss sale is a documented follow-up."*

So a mixed disposal with a nonnegative aggregate reports **zero wash-sale exposure while loss shares exist** — and my criterion would have surfaced that across all five relief methods as decision support. That is precisely the "presents a known-partial number as a settled one" failure this slate exists to prevent, which makes it worse coming from `W10-TAX-001` than from anywhere else.

Per-loss-lot decomposition with regression evidence is now a prerequisite, and there's an explicit criterion that until it lands, mixed disposals render as **incomplete rather than zero**.

### The other P1

**Elimination pairing.** Confirmed: `LedgerLineDimensionSet` carries `EntityId` and `CounterpartyId` as separate optional dimensions, and the consolidated book key has no legal-entity identity. Counterparty alone would merge balances from unrelated entities facing the same affiliate — drafting entries between the wrong entities while concealing genuine unmatched positions. Now requires reciprocal `(EntityId, CounterpartyId)` pairs scoped by account and currency.

### The four P2s

- **Recurring-journal dependencies** — right. A schedule holds only a `TemplateId` while the template book and locked-period book are separate in-memory fields, so persisting schedules and idempotency state alone leaves a restarted worker unable to materialize or evaluate its lock. Durable template resolution *with version* and an authoritative period-lock source are now required.
- **Replay repositories already exist** — right, and worth conceding plainly: `IStatementNormalizedEntityRepository` and `IStatementMatchResultRepository` have file-backed implementations and **zero references outside `StatementRepositories.cs`**. My "nothing retained supports replay" premise was wrong in exactly the way this slate's own thesis predicts. Recast as activating and extending them rather than building a second storage path — this is a seventh instance of built-but-unwired capability, which strengthens the slate's argument rather than undermining it.
- **Durable match attribution** — right, my "no contract change" claim was wrong. Persistence does `ToleranceRuleId = match.RuleIds.FirstOrDefault()` in two places: first element only, list discarded, no promoter field. Attribution lived solely in the ephemeral engine result. Now requires durable records carrying the full rule list and the promoting operator.
- **Queue default** — right. From the second run onward most unresolved work is "persisting," so a new-only default hides aging and SLA-pressured breaks behind an apparently empty queue, contradicting the very next criterion. *New* is now a marker on a row, not a filter.

### Also in this push

`main` is merged into this branch, so `verify-dotnet` now runs against #2549's fix rather than the pre-merge base. Validation: registry 0 errors, `check_program_state_consistency` passed, `verify-docs` exit 0, 39 rows with all eleven `W10-*` rows `planned` / `green`, orphan count 258.

Fifteen findings across three rounds, all valid. The consistent root cause on my side is asserting a current-state premise from a grep instead of reading the type that owns the behavior — worth naming, since this slate's entire value is being an accurate planning source.

---

rodoHasArrived:
Round five verified — all five valid, three P1 — and addressed in `10e28247b`, together with a structural change to the rows themselves.

### Why the rows changed shape

Five review rounds produced **29 findings with none rejected**, and they were not converging. The cause turned out to be the shape of the rows rather than any individual claim.

The W9 rows this slate was modelled on carry **3–6 outcome-shaped exit criteria**. Mine had grown to **7–12** and had become mechanism-shaped — naming keys, guards, storage shapes, decomposition strategies. Design in a `planned` registry row attracts design review, and that review can't terminate at the registry level, because the design hasn't been worked through against the code. That's what `meridian-blueprint` is for.

So every row now states an outcome and its acceptance evidence, **4–6 criteria each**. The round-five findings are folded in as outcomes rather than mechanisms:

| Finding | Now expressed as |
|---|---|
| Future-dated marks pass the freshness gate | "A valuation cannot rest on a mark whose age **or observation date** falls outside policy" |
| Tax policy not effective-dated | "Reopening or regenerating an earlier period reproduces the tax figures originally reported, including after a relief-policy change" |
| TWR ignores external flows | "Reported returns are correct across periods containing external capital flows" |
| Elimination drafts not idempotent | "Rerunning the same perimeter and as-of date cannot produce a duplicate elimination for the same pair" |

### Nothing found in review is lost

The slate doc gains a **Known Constraints from Source** section — per row, every verified finding, each attributed to the type that owns the behavior. The aggregate-only wash-sale deferral and the throwing non-effective-dated relief policy, the two incompatible break identifiers, the unwired bulk casework and retention repositories, the ephemeral match attribution keeping only the first rule ID, the missing pricing series and residual value, the separate entity and counterparty dimensions, and the rest.

That's blueprint input, and it's where design belongs. It opens with a note to read it before blueprinting any row, because several of these are the reason a row's outcome is phrased the way it is.

### Also in this change

- **`W10-MARK-001` gains `SRC-WPF`** — correct, the criterion promises mark age wherever positions appear and the desktop lane has `PositionBlotterViewModel` and `FundLedgerViewModel`.
- **Production-readiness tracker refreshed.** It reported "1 planned" naming only `W6-BTSTUDIO-001`. Worth noting it was **already stale for the nine open W9 rows before this PR** — my slate doc's tracker mapping just made that visible. Now reports the true count of 22 across `W5X-OEG-001`, `W6-BTSTUDIO-001`, W9, and W10, with the reviewed date advanced.

### On the tax-lot finding specifically

Verified and worse than described. `LedgerAccountTaxLotPolicyBook` keys one policy per account and `Register` replaces it, and `Resolve(account, asOf)` **throws `InvalidOperationException`** when the stored policy postdates the disposal. So after a relief-method change, regenerating an earlier report pack doesn't merely lose the original method — it fails outright, or silently produces different figures if the new policy is backdated. That's a reproducibility break on a governed artifact, and it's now captured as a constraint with the row demanding reproducibility as the outcome.

### Status

`verify-dotnet` and `quality-gate` both **passed** on `1f2af9419` before this push, confirming the base fix from #2549 works here. Validation on `10e28247b`: registry 0 errors, `check_program_state_consistency` passed, `verify-docs` exit 0, 39 rows, all eleven `W10-*` `planned` / `green`, orphan count 258.

---

rodoHasArrived:
## Round six applied in `4b71fd4ea` — and a wrong prediction to retract

Last round I said trimming the rows to outcome-shaped criteria "should end the review cycle, since the rows no longer present design for review." That was wrong. Round six arrived against the trimmed rows with eight findings, and all eight are valid. The shape was not the cause of the review volume.

Running total across six rounds: **37 findings, zero rejected.**

### Three missing source-module routes

Objectively wrong, verified against the owning file:

| Row | Added | Why |
| --- | --- | --- |
| `W10-RECON-004` | `SRC-DOMAIN` | `ReconciliationMatchLink` — the retained contract that must gain rule and promoter attribution — is in `src/Meridian.Domain/Reconciliation/StatementReconciliationAggregate.cs:35`. It carries one `ToleranceRuleId` and no promoter. A trace through Financial Operations, Strategies, and the dashboard cannot widen that schema. |
| `W10-TAX-001` | `SRC-WPF` | The desktop lane presents the same positions through `PositionBlotterViewModel` and `FundLedgerViewModel`. The final criterion now reads "operator surfaces on both workstation lanes." |
| `W10-CONSOL-001` | `SRC-DESIGN-FINANCIAL-OPERATIONS` | `AccountingPolicyService` and `AccountingJournalDraftService` are both in `src/Meridian.FinancialOperations/Ledger/` — they own the two seams the row explicitly requires. |

### Five specification gaps, closed as outcomes

Each is expressible without describing mechanism, so all five went into the rows rather than being deferred to the doc. Every row stays inside the 3–6 criteria band W9 uses.

**P1 — bulk execution not bound to the preview (`W10-RECON-002`).** Spot-checked and confirmed. `ReconciliationBulkCaseworkRequest` (`ReconciliationDtos.cs:653`) carries `BreakIds, Action, Actor, CommandId, CorrelationId, Source, IdempotencyKey, DryRun, AllowPartialSuccess, Reason?, Assignee?, Priority?` — no per-break expected versions, no preview receipt — and the repository reads each `item.Version` only once execution begins. The dry run and the execution are independent reads, so a concurrent edit between them applies the action to state the preview never showed. The criterion now requires the action to apply only to what the preview showed, in the state it showed it, with drift returning the operator to a fresh preview.

**P1 — readiness contributor absence is not fail-closed (`W10-SEAM-001`).** Spot-checked and confirmed, and it is worse than routing. `FinancialOperationsCommandCenterReadService` takes `IOperationsCloseCalendarService? closeCalendarService = null` and `IPrivateCapitalCloseCockpitService? privateCapitalCloseCockpitService = null`. Contributors are **optional constructor parameters defaulted to null**, so an unregistered or failing lane is indistinguishable from a lane with nothing to report. Consolidating onto it as written would carry the false-ready path forward rather than close it. New criterion: an unregistered, failing, stale, or out-of-scope contributor makes the projection incomplete and blocking.

**P1 — unposted elimination drafts could reach reported figures (`W10-CONSOL-001`).** Subtracting a proposed draft satisfies "no double-counting" while concealing a genuine intercompany balance behind work nobody approved. Criterion 1 now requires eliminated figures to reflect only approved and posted eliminations, with drafts visible solely as a labeled preview.

**P1 — replay retains only half a run (`W10-RECON-003`).** Verified: `StatementNormalizedEntities` (`StatementRepositories.cs:143`) holds `Positions, CashBalances, Transactions, Securities, SourceRows` — all statement-side — while `StatementMatchingEngine` also consumes `InternalPositions`, `InternalCashBalances`, and `InternalLedgerTransactions`. Replaying against live internal state would attribute intervening portfolio, cash, or ledger movement to the tolerance change, and would still *look* deterministic. Merged into the determinism criterion so the count stays at six.

**P2 — lineage conflated with occurrence (`W10-RECON-001`).** A cleared-then-recurring break could reopen its original item and keep the original SLA age, or overwrite the interval during which it was clear. New criterion requires lineage and occurrence to be distinguishable.

### What I think this actually shows

Six rounds have not converged, and the shape change did not stop it. The rows are no longer the problem — they are outcome-shaped and match W9 house style. What the reviewer is doing is adversarial specification review, and that terminates against a build, not against a registry row. A `planned` row states what "done" means and what evidence proves it; it cannot be made airtight against a reviewer asking how each outcome could be satisfied incorrectly. That question is worth asking — it produced two genuine reproducibility constraints in shipped code — but its answer belongs in a blueprint.

So the findings keep going to the right place. Every verified source fact from all six rounds is in `docs/product/w10-depth-slate-2026-07.md` under **Known Constraints from Source**, attributed to the type that owns the behavior. That section is now the most useful artifact in this PR and the place an implementer should start.

### Validation

```
python3 build/scripts/docs/validate-roadmap-registry.py --summary   # 0 errors, 0 warnings
python3 build/scripts/docs/render-roadmap-docs.py --summary
python3 build/scripts/docs/render-roadmap-diagrams.py --summary
python3 build/scripts/docs/run-docs-automation.py --profile core ...
python3 scripts/check_program_state_consistency.py                  # passed
bash scripts/ci.sh --lane verify-docs                               # exit 0, 0 drift
```

39 rows, all eleven `W10-*` `planned` / `green`, criteria counts 5–6 per row.

`main` now carries #2550, so **the lexicographic diagram finding on this PR is resolved** — W10 renders after W9-ASSET-010 rather than between W1 and W2. Within W10 it orders by area, not slate rank, which is the limitation #2550 documented: W10 identifiers deliberately do not encode rank the way W9's do. Recovering it needs an optional `sequence` schema property, tracked separately.

---

rodoHasArrived:
## Round seven applied in `0d3c9c539`

Three findings, all verified against the owning type, all valid. **40 findings across seven rounds, zero rejected.**

**P2 — `W10-PERF-001` missing `SRC-DESIGN-FINANCIAL-OPERATIONS`.** `PrivateCapitalActivityProjectionBuilder.cs` is in `src/Meridian.FinancialOperations/PrivateCapital/`, and it owns investor cash-flow dates, investor identity, and posted state. A kernel built from Ledger and Contracts alone would bypass the authoritative investor cash-flow source. Added.

**P1 — currency treatment undefined (`W10-PERF-001`).** Confirmed and sharper than reported. `BuildCapitalAccounts` groups subledger entries by `{ CapitalAccountId, InvestorId, Currency }` — currency is part of the grouping key, so an investor transacting in two currencies has **two activity rows**, and a naive kernel would sum their nominal amounts into one cash-flow series. That produces a USD contribution netted against a EUR distribution: a figure with no meaning that still satisfies every completeness label the row already requires. New criterion requires per-currency reporting or translation at each flow's and the terminal value's as-of rate, and a missing rate now counts as an incomplete input in the pro-forma criterion.

To hold the row at six criteria, the standalone golden-file criterion folded into the evidence criterion that already named it — no coverage lost, the requirement is stated once instead of twice.

**P1 — tolerance commit not bound to the previewed version (`W10-RECON-003`).** Confirmed. `StatementToleranceProfile` carries `int Version` (`StatementToleranceProfiles.cs:3`), but `FileStatementToleranceProfileProvider` is a **`static class` with no save, write, or compare-and-swap path** — load-only. Adding an edit surface without a version boundary lets a preview taken against version N be retained as justification for a commit that lands on N+1, so the applied thresholds are not the ones whose break impact was reviewed.

This is the same defect as the bulk-casework preview gap fixed last round, in a different seam. Worth naming as a pattern rather than two isolated findings: **anywhere this slate introduces a preview, the preview and the act it justifies are currently independent reads.** Both rows now require the acting side to bind to what the preview showed.

### Validation

`validate-roadmap-registry.py` 0/0 · `check_program_state_consistency.py` passed · `bash scripts/ci.sh --lane verify-docs` exit 0, 0 drift. 39 rows, all eleven `W10-*` `planned` / `green`, criteria counts 5–6 per row.

---

rodoHasArrived:
## Round eight applied in `2120c94c6`, plus a `verify-dotnet` diagnosis

### `verify-dotnet` failure on `0d3c9c539` — two flaky tests in `main`, not this PR

This branch is **docs-only** relative to `main` (`git diff origin/main...HEAD --name-only` returns nothing outside `docs/`), so nothing here can reach compiled source. Two tests failed:

**1. `Meridian.Tests.Storage.MaintenanceSchedulerTests.ScheduledJob_WhenRunningAdmissionCannotBeRetained_IsObservedAndRequeued`**
`Expected history.RunningAppendAttempts to be greater than 0, but found 0.` — test duration **25 ms**.

The wait loop polls up to 100 × 20 ms for the job to appear in `GetPendingJobs()`, then asserts the counter. But the job is placed in pending by `ScheduleAsync` *before* `Start()` is ever called, so the loop's exit condition is already true on its first iteration — it cannot distinguish "still in its initial scheduled state" from "requeued after a failed running admission." A 25 ms duration means it fell straight through and asserted on a counter the scheduler hadn't touched yet. The condition needs to wait on the admission attempt, not on pending membership.

**2. `Meridian.Tests.Execution.OrderManagementSystemReportStreamTests.UnsolicitedAcceptedModification_CannotIncreaseAuthorizedQuantity`**
`KeyNotFoundException: The given key 'AAPL' was not present in the dictionary.` — test duration **1 ms**.

It awaits `WaitUntilAsync(... Status == OrderStatus.Filled)` and then immediately reads `portfolio.Positions["AAPL"]`. Order-status transition and portfolio application are not the same atomic step, so there is a window where the status is `Filled` and the position has not been written yet.

**Both are races, not regressions.** `verify-dotnet` passed on this same branch at `1f2af9419` (23:11), and the only source-tree change between then and `0d3c9c539` was #2550 — Python and generated docs, zero C#. No compiled code changed between a green run and a red one.

**I am not fixing them in this PR.** They are async test races in the maintenance scheduler and the OMS report stream, entirely outside a registry-data change. This PR is `phase:PR1` (`docs/**` only), so touching `tests/**` would fail `scope-gate` and would break this PR's own "no unrelated changes" safety claim. They deserve their own branch. The push above re-runs CI, which will also confirm the flakiness reading.

### Round eight — five findings, all valid

**P1 — instrument identity omitted from promoted rules (`W10-RECON-004`).** Last round I tightened this criterion to protect record kind and currency. That was not enough, and the reviewer is right about why. `SamePositionIdentity` requires `SecurityId` on **both** the exact and tolerance tiers, and `SameTransactionInstrumentOrCurrency` compares `SecurityId` **instead of** currency whenever either side is security-backed. Instrument is an identity predicate in its own right, not a subcase of currency — a rule generalized on counterparty, description, account, sign, and a numeric tolerance could match an unrelated security in the same account. The criterion now requires every immutable predicate the match kind enforces.

**P2 — SLA calendars ignored (`W10-RECON-001`).** Confirmed: `ReconciliationSlaPolicy` exposes `BusinessCalendarId` and `HolidayCalendarIds` (`ReconciliationDtos.cs:579-580`) while `ReconciliationSlaCalculator` inspects only `DayOfWeek` and skips Saturday and Sunday. I had this in the constraints doc, but the row's escalation criterion did not require it — so a break around a market holiday escalates early and bands at the wrong business age. Age is now measured against the named calendars, with calendar-boundary tests in the evidence criterion.

**Three routing gaps**, each confirmed by file location:

| Row | Added | Owning file |
| --- | --- | --- |
| `W10-MARK-001` | `SRC-CONTRACTS` | `FundPortfolioPosition`, `PortfolioPositionSummary`, `WorkstationTradingPositionRow` all live in `src/Meridian.Contracts/Workstation/` and none carries a mark observation date or age — the trading row exposes `MarkPrice` with no date attached |
| `W10-RECON-001` | `SRC-WPF` | `src/Meridian.Wpf/Services/FundReconciliationWorkbenchService.cs` maps the same break-queue collection into the desktop Fund Ledger queue |
| `W10-RECON-002` | `SRC-CONTRACTS`, `SRC-UI-SHARED` | request defined in `Meridian.Contracts/Workstation/ReconciliationDtos.cs`, endpoints in `Meridian.Ui.Shared/Endpoints/WorkstationEndpoints.Reconciliation.cs` |

Both criteria edits amended existing text rather than appending, so all eleven rows stay at 5–6 criteria.

**45 findings across eight rounds, zero rejected.**

### Validation

`validate-roadmap-registry.py` 0/0 · `check_program_state_consistency.py` passed · `bash scripts/ci.sh --lane verify-docs` exit 0, 0 drift.

---

rodoHasArrived:
## Round nine applied in `06804b7ad`, plus the flakiness call confirmed

### `verify-dotnet` — flakiness diagnosis confirmed

`2120c94c6` passed `verify-dotnet` and `quality-gate`. Same C# source as the run that failed on `0d3c9c539` — this branch touches nothing outside `docs/**`, so not a byte of compiled code differed between the red run and the green one. Both failures were races, as diagnosed:

- `MaintenanceSchedulerTests.ScheduledJob_WhenRunningAdmissionCannotBeRetained_IsObservedAndRequeued` — wait loop's exit condition is satisfied by the initial scheduled state
- `OrderManagementSystemReportStreamTests.UnsolicitedAcceptedModification_CannotIncreaseAuthorizedQuantity` — status transition and portfolio application are not atomic

Still worth their own branch. They will resurface.

### Round nine — six findings, all valid

**P1 — promotion is an ungoverned straight-through lane (`W10-RECON-004`).** The strongest finding in nine rounds, and it is a gap against this repo's own charter rather than against the code. Design charter §21 permits moving from per-item to per-policy approval **only under all of**: deterministic rules rather than model output; a human-approved, versioned policy defining the eligible class with materiality caps; full retained evidence and audit events that stay reversible through governed correction; sampling review and a kill switch; and material or high-risk classes remaining per-item.

My gate required adjudicated outcomes, a sample size, and a precision bound — roughly one of those five. And the consequence is worse than under-specification: an over-broad promoted rule can suppress a material break *before* the independent-approval control ever runs. The criterion now names the charter conditions.

**P1 — XIRR returns an arbitrary root (`W10-PERF-001`).** Confirmed in `XirrCalculator`. Newton-Raphson runs from a fixed `0.10` guess and returns the first rate it converges to; the bisection fallback searches `[-0.999, 10.0]` and returns `NaN` when the endpoints share a sign. A series with several sign changes — contribution, distribution, later recall — can admit more than one valid rate, and the kernel returns whichever it reaches with nothing marking the result ambiguous. My completeness criterion blocked only *missing* inputs, so an arbitrary root would have been reported as complete. Non-convergence and multiple roots are now review-required under a declared convention.

**P1 — provenance matches on text, not identity (`W10-PROV-001`).** `FindRelatedCasesAsync` scans the tenant and company queue and substring-matches break ID, routing text, explainability text, account name, and symbol, with no enforcement of the report's fund, ledger book, period, or an exact evidence identifier. Two funds with overlapping account names or symbols in one tenant attach each other's cases to an amount — and the passport presents that as proof. The criterion now requires stable subject and evidence identifiers scoped to fund, book, and period.

**P1 — restart gate too narrow (`W10-JRNL-001`).** The owning service keeps its journal template book and locked-period book in separate in-memory fields, both of which the due calculation needs. Durable schedules alone would let a restarted worker lose a template or, worse, treat a locked period as open. Both are now inside the restart gate.

**P2 — `W10-RECON-002` gains `SRC-WPF`.** `FundReconciliationWorkbenchService` loads the same break queue and `FundLedgerViewModel.Reconciliation` exposes individual casework actions, so bulk resolution shipped to the browser alone leaves desktop operators resolving grouped breaks one at a time.

### P2 — the diagram, now raised by both reviewers

Copilot raised this independently, and both of you are right that the artifact publishes a contradictory sequence: it places rank-11 `W10-CONSOL-001` first and draws an arrow from it into rank-5 `W10-JRNL-001`.

There is an in-scope workaround — renumber the W10 identifiers so their suffix encodes rank the way W9's does. The renderer already sorts by sequence before area (#2550), so that alone would produce the adopted order with no code change, entirely inside `docs/**`.

**I am deliberately not doing that**, because it fixes the appearance and not the defect. Those edges are **synthetic for every wave**: the renderer chains all items linearly regardless of actual dependencies. W9 looks correct only because its numbering coincides with its rank. Renumbering W10 would leave the diagram asserting fabricated dependencies everywhere — including *within* W10, where the slate says ranks 5 and 6 are parallel and rank 3 is load-bearing for four separate rows. It would also bake rank into identifiers that `decision-log.yml`, `risk-register.yml`, `document-index.yml`, and the slate doc all reference, so any future re-rank would churn every one of them.

The defect is in generation: the renderer has no rank input, and it draws sequencing arrows it has no basis for. Both belong in a generator PR — the same reason #2550 was split out of this one. Happy to open it.

### Validation

`validate-roadmap-registry.py` 0/0 · `check_program_state_consistency.py` passed · `bash scripts/ci.sh --lane verify-docs` exit 0, 0 drift. 39 rows, all eleven `W10-*` `planned` / `green`, criteria counts 5–6 per row.

**54 review comments across nine rounds, zero rejected.**

---

rodoHasArrived:
## Round ten applied in `208353746`

`verify-dotnet` passed on `06804b7ad` before this push, so the round-nine state was green.

Six findings, all valid. **Two are violations of this repo's own design charter rather than mismatches with code**, which is the pattern worth noting from this round.

**P1 — consolidating mark policy could delete the stricter control (`W10-MARK-001`).** `MarkPriceQualityPolicy` (`DailyMarkToMarketService.cs:60`) governs `maximumAge`, `minimumConfidence`, `RequireCompleteCoverage`, and `RequireObservedDate`. `StalePricePolicy` (`StalePricePolicy.cs:26`) governs `Enabled`, `MaxAgeDays`, `Handling` — age only. My criterion asked for "one policy owner rather than two independently configured controls," which is satisfiable by keeping the weaker one and admitting low-confidence marks that are rejected today. It now requires the non-age gates to survive consolidation.

**P1 — a stale-mark override could become permanent (`W10-MARK-001`).** Charter §11.5 Override Strategy is explicit: *"Review or expire override — for prices, ratings, assumptions, temporary corrections."* My criterion required only that an override be dated, approved, and retained, so one approval could carry forward across later valuations while the mark grew staler. Overrides are now bound to the position, mark observation, valuation date, and policy version they were approved for, and must expire or be re-reviewed.

**P1 — pairing on counterparty alone (`W10-CONSOL-001`).** `LedgerLineDimensionSet` holds `EntityId` and `CounterpartyId` as independent optional fields. The slate doc already recorded this risk; the criterion did not require guarding against it. Pairing is now an exact reciprocal match on both, with a missing or ambiguous dimension blocking the draft.

**P1 — no consolidation currency (`W10-CONSOL-001`).** Each `LedgerBookDto` carries its own `BaseCurrency` and entries store functional amounts, so comparing them directly either nets unlike units or reports a currency gap as an intercompany mismatch — which would then surface in the very break class the row promises. Elimination now happens in one declared consolidation currency at governed as-of rates, translation differences retained separately, missing rate blocking the run.

**P2 — `W10-JRNL-001` gains `SRC-WPF`.** `AccountingConfigureViewModel.ManualJournalDraftRows` presents the same draft collection in the desktop lane.

**P2 — my error in `docs/product/README.md`.** It claimed every W9 and W10 row is planned with planned-evidence posture. `W9-ASSET-010` is `done` with `evidence_posture: complete`, so the canonical product snapshot contradicted both the registry and the tracker it links to. Corrected to the nine open W9 rows, naming the closed one.

All eleven rows stay at 5–6 outcome-shaped criteria; both criteria additions were amendments rather than appends.

### The diagram finding now has a PR

#2551 is open — it adds an optional integer `sequence` to the item schema and has `render-roadmap-diagrams.py` prefer it over the identifier suffix. The generated diagram is **byte-identical on current registry data**, so it is a pure capability addition; `scope-gate` and `regenerate-docs` are green there.

Once it lands, the W10 rows here add `sequence: 1..11` and the registry bumps to schema `1.1.0` — that bump belongs with the first data change that uses the field, not with the schema change itself.

#2551 also states the limitation this fix does **not** address: the renderer draws item-level arrows the registry has no `depends_on` data to justify, so they express ordering rather than dependency. Making them true, or removing them, changes generated output semantics — major or renderer major per `docs/roadmap/schema-versioning.md` — and is an owner decision rather than something to fold into a sort-order fix.

### Validation

`validate-roadmap-registry.py` 0/0 · `check_program_state_consistency.py` passed · `bash scripts/ci.sh --lane verify-docs` exit 0, 0 drift.

**60 review comments across ten rounds, zero rejected.**

---

rodoHasArrived:
## Round eleven applied in `782a7c4b9`

Four findings, all valid. **Two more are charter violations rather than code mismatches**, which makes five across the last three rounds — the clear pattern in this slate's remaining defects.

**P1 — eliminations not bound to source versions or a correction chain (`W10-CONSOL-001`).** The charter's ledger invariants bind this row directly:

> **Concurrency control** — Writes that depend on balance, close status, settlement status, or approval state use optimistic version checks and fail closed on stale state.
> **Posted immutability** — Posted entries cannot be edited or deleted; corrections use reversing or adjusting journals linked to the original record.

An elimination is a balance-dependent write, so it needs expected versions for **both** reciprocal sides at approval. And the same-perimeter idempotency key I added in an earlier round actively works against correction: on a corrected rerun it would preserve the obsolete elimination rather than let a new one through. Approval now fails closed when either source balance has moved since the draft was computed, and a posted elimination is corrected through a linked reversing or adjusting entry rather than replaced in place.

**P1 — tolerance commits are an ungoverned straight-through change (`W10-RECON-003`).** Last round I bound the commit to the previewed profile version, which stops preview-to-commit drift. It confers no *authority*. A widened tolerance automatically suppresses future breaks — exactly the per-item-to-per-policy move §21 governs. A successful preview establishes the effect, not the right to activate. Activation now requires versioned approval with materiality caps, retained reversible evidence, sampling review, a kill switch, and per-item handling for material or high-risk classes.

That is the second row in this slate where I fixed the mechanical defect and left the governance one standing — `W10-RECON-004` was the first. Worth noting as a pattern rather than two isolated corrections.

**P2 — `W10-CONSOL-001` gains `SRC-WPF`.** `FundLedgerReadService` builds its own consolidated trial balance through `FundLedgerScope.Consolidated`; without routing the lane it keeps presenting an unlabeled consolidated view that cannot distinguish gross from eliminated.

**P2 — `W10-RECON-004` gains `SRC-CONTRACTS`, `SRC-UI-SHARED`, `SRC-WPF`.** `MatchSelectedReconciliationItemsAsync` resolves a break through the shared reconciliation workbench service, so capture wired only to the browser would teach rules from one lane's manual matches and silently discard the other's.

All eleven rows stay at 5–6 outcome-shaped criteria; every change was an amendment rather than an append.

### Validation

`validate-roadmap-registry.py` 0/0 · `check_program_state_consistency.py` passed · `bash scripts/ci.sh --lane verify-docs` exit 0, 0 drift.

### Sequencing with #2551

#2551 bumps `roadmap-items.yml` to `meridian.roadmap-items@1.1.0` with `minimum_renderer_version: 2.0.0`, which conflicts with this branch's schema header on two lines — resolved by taking #2551's version when it lands. The W10 rows then add `sequence: 1..11`, which closes the diagram-ordering finding without any further version change, since setting the field on a row is an ordinary data update.

**64 review comments across eleven rounds, zero rejected.**

---

PR #2551, opened 2026-08-01
Share one ordering across generated roadmap views and let a wave declare explicit rank

rodoHasArrived:
## Summary

Gives every generated roadmap view one ordering — wave first, then rank within the wave — and adds an optional integer `sequence` so a wave whose identifiers do not encode rank can declare it.

**Three defects, one root cause.** `diagram_sort_key` recovered rank from the identifier's trailing number, which only works for waves that encode rank there (`DEC-PRIORITY-SLATE-001` records that W9 does). Everything else fell through to an alphabetical area sort.

1. **A wave that numbers per area has no recoverable rank.** Every row ends in `001`, so the W10 depth slate in #2548 rendered `CONSOL(11) → JRNL(5) → MARK(1) → …` against an adopted order starting at MARK and ending at CONSOL. Because the renderer chains items with `-->`, the artifact drew rank-11 work as a prerequisite for rank-5 work.

2. **Only the diagram was ever fixed.** `render_summary` and `render_register` still sorted on the raw identifier string — the same lexicographic bug #2550 fixed for the diagram, still live in two artifacts. `ROADMAP_SUMMARY.md` renders `W1-DATA-001, W10-CONSOL-001, … W10-TAX-001, W2-PROMO-001` — W10 between W1 and W2 — and renders W9 alphabetically rather than in its adopted rank. #2550 fixed one view of three.

3. **`sequence` had no enforcement.** `validate-roadmap-registry.py` checks only the schema *header* and never loads `roadmap-items.v1.schema.json`, so `minimum: 1` and `type: integer` bound nothing. `0`, `-2`, `1.5`, `"abc"`, and `true` all passed `verify-docs`, after which the renderers either took a nonpositive rank or silently fell back to the suffix — producing a mis-ordered view from data CI had declared valid.

The sort key now lives in `common.py` as `roadmap_item_sort_key`; the diagram, summary, and register all consume it, so they cannot disagree about delivery sequence. Regenerating puts W9 in its adopted order:

```
W9-TRUTH-001 → W9-DEMO-002 → W9-PAPER-003 → W9-ALPACA-004 → W9-REPORT-005
→ W9-NAV-006 → W9-SAFETY-007 → W9-GOV-008 → W9-INGEST-009 → W9-ASSET-010
```

The `.mmd` is byte-identical — #2550 already fixed it. The summary and register reorder, which is defect 2 being corrected, not incidental churn.

## Reason

Raised on #2548 by both `chatgpt-codex-connector` and `copilot-pull-request-reviewer`, and deferred there twice because a generator change does not belong in a registry-data PR — the same reason #2550 was split out.

The in-scope workaround on #2548 was to renumber the W10 identifiers so their suffix encodes rank. Rejected deliberately: it would bake rank into identifiers that `decision-log.yml`, `risk-register.yml`, `document-index.yml`, and the slate document all reference, so any future re-rank would churn every one of them. Rank is view metadata and belongs in a field.

## Versioning

Per [`docs/roadmap/schema-versioning.md`](docs/roadmap/schema-versioning.md):

| Change | Rule | Applied |
| --- | --- | --- |
| Optional `sequence` property added | Add optional field → **Minor** | registry declares `meridian.roadmap-items@1.1.0` |
| Summary and register reordered | Change generated output semantics → **Major or renderer major** | `ROADMAP_RENDERER_VERSION = "2.0.0"`, emitted in generated headers and `MANIFEST.json`; `minimum_renderer_version: 2.0.0` |

The schema bump lands here rather than with first use: the contract changed the moment `sequence` became an accepted property, and a consumer needs to distinguish the extended schema from 1.0.0 whether or not current rows exercise it. `generated_header` and `write_manifest` take the version as an optional argument defaulting to `1.0.0`, so the four other generators sharing those helpers are untouched — per-generator versioning is what makes it auditable.

## Known limitation, stated deliberately

**The renderer still draws dependency arrows it has no data for.** Items are chained in sort order, so the edges express *ordering*, not dependency — including within a wave whose planning document says some rows are parallel and one is load-bearing for four others. This change makes the ordering correct; it does not make the arrows true.

The registry has no `depends_on` field, so a real dependency graph cannot be rendered from current data. The honest options are to add one, or to keep arrows only between waves. Both change generated output semantics again, and that is a roadmap-owner decision rather than something to fold in here.

## Testing performed

- [x] `bash scripts/ci.sh --lane verify-docs` — exit 0, 0 drift
- [x] Relevant unit tests were added or updated — seven new

`python3 build/scripts/ci/run-script-tests.py` → **425 tests, 0 failures, 0 errors**.

| Test | Pins |
| --- | --- |
| `test_explicit_sequence_orders_a_wave_that_numbers_per_area` | defect 1 — MARK(1) → PROV(3) → JRNL(5) → CONSOL(11) |
| `test_explicit_sequence_overrides_the_identifier_suffix` | declared rank wins over a conflicting suffix |
| `test_items_without_sequence_still_order_by_identifier_suffix` | the field is optional; a partially annotated wave still orders sanely |
| `test_generated_roadmap_views_share_one_ordering` | defect 2 — one key across diagram, summary, register |
| `test_out_of_range_or_wrong_typed_sequence_fails_registry_validation` | defect 3 — `0`, `-2`, `abc`, `1.5`, `true` each rejected |
| `test_valid_sequence_passes_registry_validation` | a legitimate rank still validates |
| `test_wave_disagreeing_with_its_identifier_fails_registry_validation` | an item cannot sort as one wave while labelling itself another |
| existing ordering + label tests | unchanged, still passing |

`sequence` is validated as an integer ≥ 1 by a single `is_usable_sequence` predicate shared between the validator and the renderers, so the two cannot drift apart. `bool` is excluded explicitly — it is an `int` subclass, so `sequence: true` would otherwise sort as rank 1.

**Phase marker:** `phase:PR6`, declared as the HTML-comment marker at the top of this body and confirmed with `tools/roadmap/enforce_phase_scope.py`. Touches `build/**`, `docs/**`, and `tests/**`; PR6 is the narrowest phase covering all three. The first push declared the phase in prose only, which `scope-gate` correctly rejected — it requires the comment marker or a `phase:PRx` label.

## Safety review

## Governance changes

Touches `build/scripts/docs/{common,render-roadmap-diagrams,render-roadmap-docs,validate-roadmap-registry}.py`, `docs/roadmap/schemas/roadmap-items.v1.schema.json`, `docs/roadmap/roadmap-item-template.md`, the registry schema header, two test files, and regenerated outputs. No workflow, CODEOWNERS, `AGENTS.md`, or `scripts/ci.sh` changes.

## Follow-on for #2548

This bumps `roadmap-items.yml` to `1.1.0` / `2.0.0`, a two-line conflict with #2548's schema header, resolved by taking this version. #2548's W10 rows then add `sequence: 1..11`, closing the diagram-ordering finding with no further version change — setting the field on a row is an ordinary data update.

### DIFF
--- a/build/scripts/docs/common.py
+++ b/build/scripts/docs/common.py
@@ -231,6 +231,64 @@ def front_matter(markdown: str) -> dict[str, str]:
     return result
 
 
+# Roadmap identifiers are `W<wave><suffix>-<AREA>-<NNN>`, e.g. W1-DATA-001, W5X-CONNECT-001,
+# W5X-STMT-ONBOARD-001 (two-token area), W9-ASSET-010, W10-MARK-001.
+_WAVE_PATTERN = re.compile(r"^W(\d+)([A-Z]*)$")
+
+# Sorts unparsable tokens deterministically last instead of letting them collide with wave 1.
+_UNPARSABLE_SEQUENCE = 1_000_000
+
+
+def roadmap_item_sort_key(item: dict) -> tuple[int, str, int, str, str]:
+    """Order roadmap items by wave, then by rank within the wave.
+
+    Every generated roadmap view shares this ordering so the diagram, the summary, and the register
+    cannot disagree about delivery sequence.
+
+    Sorting on the raw identifier string is lexicographic, which places `W10-` between `W1-` and
+    `W2-` and orders items within a wave alphabetically by area. Both are wrong: the first renders a
+    planned wave in the middle of delivered work, and the second discards the rank that the numeric
+    suffix encodes (see `DEC-PRIORITY-SLATE-001`, which states rank is carried in each W9 item's
+    numeric suffix).
+
+    Waves sort numerically with their alphabetic suffix as a tiebreak, so `W5` precedes `W5X` and
+    `W9` precedes `W10`.
+
+    Within a wave, an explicit `sequence` wins when the item declares a usable one. Otherwise rank
+    falls back to the identifier's trailing number, which is only meaningful for waves that encode
+    rank there. A wave that numbers per area instead - every row `-001` - has no recoverable rank
+    without the explicit field. Area and identifier remain as stability tiebreaks.
+    """
+    identifier = item.get("id", "") or ""
+    parts = identifier.split("-")
+    wave_token = parts[0] if parts else ""
+    wave_match = _WAVE_PATTERN.match(wave_token)
+    wave_number = int(wave_match.group(1)) if wave_match else _UNPARSABLE_SEQUENCE
+    wave_suffix = wave_match.group(2) if wave_match else wave_token
+
+    declared = item.get("sequence")
+    if is_usable_sequence(declared):
+        sequence = declared
+    else:
+        try:
+            sequence = int(parts[-1])
+        except (IndexError, ValueError):
+            sequence = _UNPARSABLE_SEQUENCE
+
+    area = "-".join(parts[1:-1])
+    return (wave_number, wave_suffix, sequence, area, identifier)
+
+
+def is_usable_sequence(value: Any) -> bool:
+    """Whether a declared `sequence` is a rank the renderers may sort on.
+
+    `bool` is an `int` subclass in Python, so `sequence: true` would otherwise sort as rank 1. The
+    schema also bounds the field at 1; a nonpositive value is rejected here rather than silently
+    ordering ahead of every legitimate rank. Validation reports these; the renderers fall back.
+    """
+    return isinstance(value, int) and not isinstance(value, bool) and value >= 1
+
+
 def markdown_table(headers: Sequence[str], rows: Iterable[Sequence[Any]]) -> str:
     output = [
         "| " + " | ".join(headers) + " |",
--- a/build/scripts/docs/common.py
+++ b/build/scripts/docs/common.py
@@ -299,14 +299,25 @@ def markdown_table(headers: Sequence[str], rows: Iterable[Sequence[Any]]) -> str
     return "\n".join(output)
 
 
-def generated_header(generator: str, schema_versions: Sequence[str], inputs: Sequence[str]) -> str:
+def generated_header(
+    generator: str,
+    schema_versions: Sequence[str],
+    inputs: Sequence[str],
+    generator_version: str = "1.0.0",
+) -> str:
+    """Emit the standard generated-doc header.
+
+    `generator_version` is per-generator: a generator that changes its output semantics bumps its
+    own version so an audit can tell which ordering or shape contract produced a committed artifact,
+    without disturbing every other generated doc.
+    """
     schema_lines = "\n".join(f"  - {item}" for item in schema_versions)
     input_lines = "\n".join(f"  - {normalize_path(item)}" for item in inputs)
     return (
         "<!--\n"
         "generated: true\n"
         f"generator: {normalize_path(generator)}\n"
-        "generator_version: 1.0.0\n"
+        f"generator_version: {generator_version}\n"
         f"render_contract: {GENERATOR_CONTRACT}\n"
         "schema_versions:\n"
         f"{schema_lines}\n"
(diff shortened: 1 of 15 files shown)


rodoHasArrived:
## Both review findings applied in `15f7de380` — and one of them found a live defect

Copilot and Codex independently raised the validation gap; Codex additionally caught that the field was honored by only one of three generated views. Both valid, and the second turned out to be worse than reported.

### The `sequence` field was honored by only one renderer — and the other two are already wrong

`render_summary` and `render_register` both sorted on `item.get("id", "")`, the raw identifier string. That is the **same lexicographic bug #2550 fixed for the diagram, still live in two artifacts**. Checked against the W10 branch, `ROADMAP_SUMMARY.md` currently renders:

```
W1-DATA-001, W10-CONSOL-001, W10-JRNL-001, … W10-TAX-001, W2-PROMO-001, W2-TRD-001, …
```

W10 between W1 and W2. And W9 renders `ALPACA, ASSET, DEMO, GOV, INGEST, NAV, PAPER, REPORT, SAFETY, TRUTH` — alphabetical, discarding the rank `DEC-PRIORITY-SLATE-001` records in its numeric suffix. #2550 fixed one of three views; the other two kept both defects.

The sort key now lives in `common.py` as `roadmap_item_sort_key`, and all three renderers consume it, so the diagram, the summary, and the register cannot disagree about delivery sequence. Regenerating puts W9 in its adopted order:

```
W9-TRUTH-001 → W9-DEMO-002 → W9-PAPER-003 → W9-ALPACA-004 → W9-REPORT-005
→ W9-NAV-006 → W9-SAFETY-007 → W9-GOV-008 → W9-INGEST-009 → W9-ASSET-010
```

This is why the diff now touches `ROADMAP_SUMMARY.md` and `roadmap-register.md` — that is the bug being corrected, not incidental churn. The `.mmd` is still byte-identical.

### `sequence` validation was decorative

Confirmed: `validate-roadmap-registry.py` checks only the schema *header* — id and major version — and never loads `roadmap-items.v1.schema.json`. So `minimum: 1` and `type: integer` bound nothing. `sequence: 0`, `-2`, `1.5`, `"abc"`, and `true` all passed `verify-docs`, after which the renderers either accepted a nonpositive rank or silently fell back to the identifier suffix.

That is precisely the failure this PR exists to prevent — a generated view ordered against its adopted rank, from data CI had declared valid. Shipping the field without closing it would have been self-defeating. The canonical validator now rejects it, sharing a single `is_usable_sequence` predicate with the renderers so validation and rendering cannot drift apart. `bool` is excluded explicitly, since it is an `int` subclass and `sequence: true` would otherwise sort as rank 1.

### Template wording

Corrected: introducing the optional field was the one-time schema change, the registry's `schema.version` bumps once when a row first uses it, and setting `sequence` on a row afterwards is an ordinary data update. The ordering section now also names all three generated views rather than implying the diagram alone.

### Validation

`python3 build/scripts/ci/run-script-tests.py` → **424 tests, 0 failures, 0 errors** (three new).

| Test | Pins |
| --- | --- |
| `test_out_of_range_or_wrong_typed_sequence_fails_registry_validation` | `0`, `-2`, `abc`, `1.5`, `true` each rejected by the canonical validator |
| `test_valid_sequence_passes_registry_validation` | a legitimate rank still validates |
| `test_generated_roadmap_views_share_one_ordering` | the shared key orders waves numerically and honors declared rank |

`bash scripts/ci.sh --lane verify-docs` — exit 0, 0 drift.

---

rodoHasArrived:
## Round two applied in `e44f6de86`

All three findings point at `docs/roadmap/schema-versioning.md` — the document this PR quoted when arguing that changing the diagram's arrows would be renderer-major, and then failed to apply to its own changes. That inconsistency is the finding behind all three.

**Schema minor.** The registry now declares `meridian.roadmap-items@1.1.0`. My earlier reasoning — that 1.0.0 data stays valid against the extended schema, so the bump belongs with first use — confused *data validity* with *contract identity*. The contract changed the moment `sequence` became an accepted property; a consumer needs to tell the extended schema from 1.0.0 whether or not current rows exercise it. The versioning table is unambiguous: **Add optional field → Minor.**

**Renderer major.** Switching the summary and register from lexical identifier order to shared wave-then-rank order changes generated output semantics, which the same table classifies as **major or renderer major**. `render-roadmap-docs.py` now declares `ROADMAP_RENDERER_VERSION = "2.0.0"` and emits it in both the generated headers and `MANIFEST.json`:

```
generator_version: 2.0.0
schema_versions:
  - meridian.roadmap-items@1.1.0
```

`generated_header` and `write_manifest` take the version as an optional argument defaulting to `1.0.0`, so the other four generators that share those helpers are untouched — the version is per-generator, which is what makes it useful for an audit. `minimum_renderer_version` advances to `2.0.0` to match.

**Wave consistency.** The sort key parses the wave from the identifier while the diagram labels with the declared `wave` field, and nothing required them to agree — `id: W10-MARK-001` with `wave: W2` would sort among W10 work while presenting itself as W2. The canonical validator now rejects the mismatch. Verified that no current row disagrees, so this fences off a latent inconsistency rather than repairing live data.

### Validation

`python3 build/scripts/ci/run-script-tests.py` → **425 tests, 0 failures, 0 errors**. New: `test_wave_disagreeing_with_its_identifier_fails_registry_validation`.

`bash scripts/ci.sh --lane verify-docs` — exit 0, 0 drift.

### Note for #2548

This bumps `roadmap-items.yml` to `1.1.0` / `2.0.0`, which #2548 also edits. That is a two-line conflict on the schema header, resolved by taking this version. #2548's W10 rows then add `sequence: 1..11` with no further version change, since setting the field on a row is an ordinary data update.

---

PR #2555, opened 2026-08-01
Reconcile project blueprints into one coherent set

rodoHasArrived:
## Summary

Documentation-only change. No Meridian runtime behavior changes.

The repository carries nine active blueprints across four homes (`docs/engineering/blueprints/`, `docs/development/accounting-blueprints/`, `docs/product/`, `docs/plans/`) with no shared index. Several conflicted with each other, and several described work that has since shipped. This reconciles them into one mutually consistent set and gives them a single register.

> **Note:** this description covers the original reconciliation. Five subsequent review rounds (Copilot + Codex) found 29 further issues, all confirmed against source and fixed — including one design decision of mine that was wrong on its merits. The per-round detail is in the PR comments; the material changes since the first commit are folded into the tables below.

### Hard conflicts fixed

| Conflict | Resolution |
|---|---|
| **Ledger migration ordinals.** Incentive-fee claimed `V_ledger_024/025`, equalization claimed `024/025/026`, commitment claimed `024` — colliding with each other *and* with the `024`–`028` scripts already on disk. All three stated the baseline as `V_ledger_023`. | Reserved non-overlapping ranges (incentive-fee **029–030**, commitment **031–032**, equalization **033–035**), corrected the baseline to `V_ledger_028__wash_sale_activation.sql`, and recorded the reservation table centrally with a "re-derive from disk at implementation time" rule. |
| **DDL precision.** The three accounting blueprints specified `numeric(28,8)`, `(28,10)`, `(28,12)`, `(20,2)`, `(9,6)` and `(18,8)` for equivalent columns. Only `numeric(38, 12)` and `numeric(38, 10)` exist in real ledger migrations. | Converged all new subsidiary-ledger tables on `numeric(38, 12)`; reserved `numeric(38, 10)` for journal-leg columns that must match `ledger_entries`. Noted that storage precision is not the rounding policy, and that the convention governs numeric columns only. |
| **API route prefixes.** Equalization used `/api/accounting/…`, financing used `/api/accounting/…` + `/api/financing/…`. Neither prefix exists in `UiApiRoutes`. | Moved both under the shipped `/api/ledger/…` prefix, beside the `/api/ledger/incentive-fee/…` and `/api/ledger/private-capital/…` surfaces they interoperate with. Recorded the prefix convention once. |
| **High-water-mark ownership.** Incentive-fee `Fork G = InvestorSeries` assumed durable *per-investor* HWM state; equalization §9 said Method A has *no* per-investor HWM and Method B uses *per-series* HWM. The doc answering incentive-fee's open question O-3 contradicted it. | `incentive_fee_state` is now the **single durable owner** under both methods, keyed by **series, never investor**, storing `high_water_mark_per_share` (a total-NAV HWM would charge fee on subscriptions). Equalization's competing `fund_series.high_water_mark_per_share` is dropped; `PartnershipInvestorAllocationInput.HighWaterMark` is labelled a transient projector input, not an owner. Scope lifecycle (`Live`/`Closed`/`Consolidated`) and three transactional ports — create, consolidate, close — are specified, plus an enablement gate that fails closed rather than defaulting a missing HWM to zero. |
| **Stream fan-out PR D.** The fan-out blueprint still proposed `workspace` / `inbox` topics and a workspace-poller suspension that the report-run blueprint had already analysed and rejected. | Marked them **cancelled, not deferred**, and cross-linked the rescope in both directions. |
| **Terminology.** Two blueprints sharing one crystallization boundary used `crystallize` vs `crystallise` and `Equalization` vs `Equalisation` in *route segments and identifiers*. | US spelling for code, routes, and columns; UK prose tolerated. Three shipped UK-spelled members are recorded as **grandfathered exceptions** rather than renamed. Two proposed types that the normalization collided with shipped types (`EqualizationAdjustment`, `EqualizationMethod`) were renamed or converted to an append. |

### Status drift corrected against source

- **Financing** — depreciation *calculation core and governed-draft seam* shipped (`DepreciationScheduleCalculator`, `FixedAssetDepreciationProjector`, `FixedAssetDepreciationDraftBuilder` with a submittable-approval test). No store, migration, orchestrating service, endpoints, or read model; repo and borrower-side debt still design-only. Phase 2 checklist corrected from fully-ticked.
- **Commitment & capital-call** — `PrivateCapitalCommitments`, `CommitmentRollForwardCalculator`, `CapitalCallDraftFactory`/`PlanBuilder`/`ScheduleDraftBuilder` and three enum members shipped; persistence, endpoints, and workbench still design-only.
- **Equalization** — `EqualizationCalculator` ships as an **entry-exposure helper only**: it takes no `GAV_cryst`, so it is not the §5.1/§5.2 crystallization math, and its NAV parameter is the dealing NAV where §5.1 needs gross `GAV_d`.
- **Quote-stream fan-out** — now `Implemented`, with the two post-delivery divergences called out.
- **Report-run status stream** — D1/D3 as designed; **D2 partially shipped**: the shared SSE helper landed additively, so `WorkstationEndpoints.Stream.cs` still carries a duplicate loop and `ResolveStreamSessionId`.
- **Portfolio cash ladder** — the known-deviation note described a unit-quantity fallback that source explicitly refuses; corrected to the fail-closed behavior.
- **Security Master Passport Workbench** — was `Proposed` while Phases 1–4 and all three browser editor slices had shipped.

### Discoverability

- `docs/engineering/blueprints/README.md` becomes the **canonical register for every blueprint in the repository**, carrying the shared conventions and cross-blueprint contracts plus an "adding a blueprint" checklist.
- The product, engineering, plans, and accounting indexes link to it. The two web-UI stream blueprints (previously unindexed anywhere), the report-writer blueprint, and the Security Master passport workbench are now registered.
- The register also records four traps found during review: enum collisions, DTO layering (`Meridian.Contracts` is a leaf with no project references), name-normalization collisions, and HWM unit discipline.
- Added the missing `Status` / `Owner` / `Reviewed` lifecycle fields that `validate-docs-structure.py` was warning about.
- The design-system `WORKSTATION_BLUEPRINT.md` is explicitly scoped **out** of the register.

## Reason

Independently authored blueprints had drifted into direct conflict. Implementing any two of the accounting blueprints as written would have produced colliding migration filenames, three incompatible column precisions, invented route prefixes, and two competing high-water-mark stores. Separately, several blueprints presented shipped work as pending, which misdirects anyone picking one up.

## Testing performed

- [ ] `bash scripts/ci.sh` completed successfully — not run; this change touches only `docs/` and compiles nothing
- [ ] Relevant unit tests were added or updated — n/a, documentation only
- [ ] Relevant integration tests were added or updated — n/a, documentation only
- [ ] GitHub Actions `quality-gate` passed — re-running on the current head

Documentation validation run locally on the current head:

- `python3 build/scripts/docs/validate-docs-structure.py` — passes; zero warnings on any touched file
- `python3 build/scripts/docs/repair-links.py --summary` — **PASS**, 1595 links checked, 0 broken
- `python3 build/scripts/docs/run-docs-automation.py --profile core` — all steps succeed; regenerated dashboards are committed

## Reviewer note — please read before merging

These blueprints specify a fund fee engine that **nothing compiles and no test exercises**. Review across five rounds found not only propagation gaps but a load-bearing arithmetic error (a total-NAV high-water mark that would charge performance fee on investor subscriptions). The documents are now internally consistent and every claim is verified against source, but *consistent* is not *correct*.

**The Method A / Method B fee math should get a human read from someone with fund-accounting background before anyone implements against it.** That is the one thing this PR cannot self-certify.

## Safety review

## Governance changes

---
_Generated by [Claude Code](https://claude.ai/code/session_01GkRguruSqnrN6hymYf7PGf)_


rodoHasArrived:
All 14 review comments (9 Copilot, 5 Codex) collapse to seven distinct findings. I verified each against source before acting — **all seven were correct**, and all are now fixed in `c28f75a`.

Two of them asked me to *make* a decision rather than restate one, so here is which way each went.

### HWM ownership — `incentive_fee_state` is the single durable owner

Both reviewers were right that the previous contract asserted the conflict away instead of resolving it: §5.3's record still declared `HighWaterMark` keyed by `InvestorId`, and `V_ledger_030` still persisted `investor_id` / `high_water_mark`, while Method B separately persisted `fund_series.high_water_mark_per_share`.

Worth noting for context: **neither store ships today** — `grep high_water` across `src/**/*.sql` and `src/**/*.cs` returns nothing. Both were proposed, so this was a design decision the docs had to record, not a code conflict.

Resolution: `incentive_fee_state` owns the HWM under **both** methods. Fork G selects a row's *scope*, not a different store.

- Rekeyed `investor_id` → `series_id` throughout the record, DDL, index, and DTO. The scope is a **series, never an investor** — that rename alone removes the per-investor reading.
- Split the value by scope under a check constraint: `high_water_mark` (fund NAV terms, set iff `series_id is null`) vs `high_water_mark_per_share` (per-share, set iff `series_id` is set). One row can never carry both.
- **Dropped `fund_series.high_water_mark_per_share`.** `FundSeriesDto.HighWaterMarkPerShare` survives as an annotated joined read.
- Codex's catch that `PartnershipInvestorAllocationInput.HighWaterMark` "is only a transient input, not a store" is exactly right — it is a `sealed record` constructor parameter. It is now labelled a projector input hydrated from `incentive_fee_state`, and the §5.4 roller is documented as the only writer.

### `EqualizationCalculator` — entry-exposure helper, not a base to build on

Codex's P1 was correct and the arithmetic checks out. `Compute(navPerUnit, highWaterNavPerUnit, subscriptionUnits, performanceFeeRate)` takes no `GAV_cryst`, so it can produce neither §5.1's bounded `creditReturned` nor §5.2's recovery-to-date redemption. And its parameter is the dealing NAV where §5.1 needs gross `GAV_d`: the §13.1 golden case returns `0.20 × (116 − 100) × 100 = 320` against a required `400`.

The "build the lot-level projection **on top of** that calculator rather than re-deriving the arithmetic" instruction was therefore a trap. Retracted in the blueprint and both indexes.

### The other five

| Finding | Fix |
|---|---|
| Financing Phase 2 fully ticked, partially shipped | Confirmed `FixedAssetRecordDto`, `IFixedAssetRegisterStore`, `FixedAssetDepreciationService`, the routes, `types.ts` DTOs, and the read model are all absent. Unticked; what landed vs. what remains now stated explicitly. Note the domain types (`DepreciationMethod`, `DepreciationPeriod`) ship without their `*Dto` counterparts — called out, since it's an easy mis-read. |
| Spelling convention asserted, not applied | Normalised every *proposed* identifier — types, interfaces, DTOs, enum members, tables, columns, indexes, and the `--filter` test selector — to US spelling; prose keeps UK. |
| Shipped UK-spelled members | Recorded as **grandfathered exceptions** rather than renamed: `EqualizationAdjustment.EqualisationCredit`, `.HasEqualisationCredit`, and `ShareClassUnitRegisterProjector`'s field. The type is US while its members are UK — that split is in source, and the register now says so instead of claiming the convention already holds. |
| Register omitted `security-master-passport-workbench.md` | Added. Its header still said `Proposed (blueprint)` while Phases 1–4 are `[x]`; reconciled to "Largely implemented" with the five open `[~]` items named. Also recorded `/api/security-master/` as a shipped route prefix. |
| D2 "shipped" | `WorkstationEndpoints.Stream.cs` has **zero** references to `StreamEndpointHelpers` and still carries its own heartbeat constant, SSE loop, and `ResolveStreamSessionId`. The extraction was additive, not a refactor. Marked partially shipped, divergence recorded in the blueprint and the register. |

One correction to Copilot's framing on the last row: it read as one leftover helper, but the file shares *nothing* with `StreamEndpointHelpers` — so it's two full SSE loops, which is the duplication D2 existed to remove.

**Validation:** `repair-links` 1593 links / 0 broken; `validate-docs-structure` clean on every touched file; docs automation `core` profile regenerated so `regenerate-docs` stays green.

---

rodoHasArrived:
Both correct, both fixed in `d6cd3af`. Checking the second one systematically turned up a third instance the review didn't flag.

### P1 — per-share vs total-NAV unit mixing

Confirmed. §6.1 defines `HWM_s = P0` as a **per-share** value while scaling `BeginningNav`/`EndingNavBeforeFees` to series *totals*, and `PartnershipInvestorAccountingProjector` computes `Max(0, EndingNavBeforeFees − HighWaterMark − managementFee)` — a total-NAV subtraction. Your arithmetic checks out: 100 shares at 110 with `HWM_s = 100` at 20% gives `0.20 × (11,000 − 100) = 2,180` against a correct `0.20 × (11,000 − 10,000) = 200`.

Fixed by specifying the conversion on both sides of the call rather than leaving it implied:

```
HighWaterMark = HWM_s × units_s                              // scale in
HWM_s'        = projection.UpdatedHighWaterMark / units_s    // scale out, then persist
```

Recorded in §6.1, repeated at the §9 instruction site where the projector call is actually specified, and mirrored in the incentive-fee roller since that's what writes the per-share column. Also noted that scale-in and scale-out must use the *same* `units_s` or the crystallised HWM won't round-trip.

For the record on ownership: this was latent in §6.1, but the per-share/absolute column split I added in `c28f75a` is what made it concrete. Mine either way.

### P2 — `EqualizationAdjustment` collision

Correct, and worth naming the cause: **my own US-spelling normalization created it.** The proposed record was `EqualisationAdjustment` (UK) before `c28f75a`, which didn't collide with the shipped `EqualizationAdjustment`. Normalizing the spelling collapsed two distinct names into one. Renamed to `EqualizationLotAdjustment` (and `EqualizationLotAdjustmentDto`).

### Third instance — `EqualizationMethod`, not reported

Since that failure mode is mechanical, I extracted every `public enum|record|interface|class` the blueprint declares and grepped `src/` for each. One more hit, same root cause:

`Meridian.Ledger.EqualizationMethod` already ships in `ShareClass.cs` as `{ None = 0, Equalisation = 1 }`, is carried on `ShareClass.EqualizationMethod`, and gates the Method A branch in `ShareClassUnitRegisterProjector`. The blueprint declared a second enum of that name and namespace with different members.

Rather than rename it, the blueprint now **appends to the shipped enum** — ordinal 1 already means exactly this blueprint's "equalisation credit / debit" method, so `SeriesOfShares = 2` is the only new member. Member 1 keeps its UK spelling because the projector compares against it; that's recorded as a grandfathered exception, not an oversight. Fork 1's option table, the `equalization_method` DDL comment, and the implementation checklist all follow the shipped names now.

I ran the same sweep across both sibling accounting blueprints. The commitment blueprint's hits are its own shipped types and are already flagged built with "do not re-derive" guidance — no action. The register now carries two preventions: grep `src/` for `enum X` before declaring one, and check for a shipped US-spelled type before normalising a UK name to it.

**Validation:** `repair-links` 1594 links / 0 broken; `validate-docs-structure` clean on every touched file; fence count unchanged from `HEAD`; docs automation `core` profile regenerated.

---

rodoHasArrived:
All four correct, fixed in `ab3b30c`. Three of them are consequences of my own previous two commits, which is the pattern worth naming — see the bottom.

### P1 — series scope stopped at the record and DDL

Right. `GetIncentiveFeeStateAsync(Guid, string? investorId, …)` was untouched, the period request still carried an investor roster, and §5.3's title, the workbench, and the test plan were all still per-investor. Following those sections would query the removed `investor_id` column, or — worse, because it fails silently — collapse two series held by the same investor onto one HWM.

- Store: `seriesId`, plus `ListIncentiveFeeStatesAsync(ledgerBookId)` to enumerate scopes (one row under Method A, one per series under Method B).
- Request: a **series roster** carrying each series' units outstanding, since the HWM is per-share and the projector needs total-NAV terms.
- Test 22 was `InvestorSeries_ComputesFeePerInvestorHwm` — now `…PerSeriesHwm`, and it asserts the collapse case directly: two series belonging to the *same* investor must price differently.

### P1 — nothing wrote the series HWM row

This is the sharpest of the four, and it's a hole I opened. Dropping `fund_series.high_water_mark_per_share` left `FundSeriesDefinition.HighWaterMarkPerShare` as an in-memory seed with no durable writer anywhere in the checklist. A new series would lose its protected level on reload, and the next crystallization would hydrate nothing — charging performance fee on the entire series NAV rather than on gains above issue price.

Series creation is now specified as a **two-store transaction**: `IFundSeriesStore.UpsertAsync` plus an `incentive_fee_state` row seeded at `IssuePrice`, committing together or not at all. With the symmetric close on consolidation and a join on load, and the field annotated as a creation-time seed that must never be read as current state.

### P2 — DTO layering

Confirmed, and the graph is even flatter than described: **`Meridian.Contracts` has no `ProjectReference` at all.** It's a leaf — `Meridian.Ledger` → `Meridian.Core` → `Meridian.Contracts`. Typing a Contracts DTO with `Meridian.Ledger.EqualizationMethod` needs Contracts→Ledger and inverts that. Contracts now owns US-spelled wire enums mapped at the service boundary, which also keeps the grandfathered `Equalisation` member off the public contract — the two problems have one fix.

The other two `EqualizationMethod` uses are fine: they're in `Meridian.Ledger` and `Meridian.FinancialOperations.Ledger`, both of which legitimately reference the domain enum.

### P2 — stale cash-ladder fallback

Confirmed verbatim. Source reads *"Active Security Master subjects at unit quantity are not a financial fact and are never used as a fallback"* and returns `([], ["No authoritative holdings source is registered; Security Master subjects cannot substitute for held quantities."])`, which `BuildAsync` also raises as a decision blocker. Corrected to the fail-closed behavior, with the superseded claim noted so the change stays traceable.

---

**On the pattern:** three of these four existed only because of my previous commits — the HWM rekey didn't propagate, dropping a column left no writer, and normalizing spelling collided with a shipped type. Each round has found fewer and narrower issues, but the shape is consistent: **a change to the contract that isn't followed to its dependent surfaces.** The register now carries four preventions (enum collision, DTO layering, name-normalization collision, HWM unit discipline).

Worth stating plainly, since this PR proposes a lot of C# and SQL that nothing compiles: the review loop is doing the work a type-checker would. I've added a repeatable collision sweep — extract declared types, grep `src/` for each — but nothing runs it in CI, so it's only as good as remembering to run it.

**Validation:** `repair-links` 1594 links / 0 broken; `validate-docs-structure` clean on every touched file; fence counts unchanged from `HEAD`; docs automation `core` profile regenerated.

---

rodoHasArrived:
All six correct. Fixed in `a237d67`. The first one **invalidated a design decision I made two commits ago**, so I've reversed it rather than patched around it.

### P1 — Method A HWM in total-NAV terms was wrong

This is the one that matters. Your example is exact: a fund at HWM `10,000` with 100 shares that issues 100 more at `100` has ending NAV `20,000` and zero market return, and the projector computes `0.20 × (20,000 − 10,000) = 2,000` of fee on pure contributed capital. And equalisation *cannot* correct it — that step preserves the projector's total fund fee and only redistributes it, which is exactly the invariant §11 asserts.

Two things make it worse than a corner case. Mid-period subscriptions are the **primary scenario this blueprint exists for**, so the flaw sat on the main path. And it contradicted this blueprint's own glossary (§3), which has always defined `HWM` as *"Fund high-water-mark **per share**"* — I introduced the total-NAV framing in `c28f75a` and didn't check it against the document's own definitions.

Rather than add a capital-flow adjustment, I reversed the decision: **both methods now store `high_water_mark_per_share`**, and every projector call scales in (`× unitsOutstanding`) and back out. That deletes the two-column split and its check constraint, and collapses Method A and Method B into one rule instead of two. The schema is simpler than before the fix.

### The three gaps that decision opened

| Finding | Fix |
|---|---|
| **Zero units (P1)** — full redemption on a crystallization date makes the scale-out divide by zero, and crystallize-on-redemption is explicitly supported. | Price on **pre-redemption** units, write no per-share HWM, close the scope. A later series for the same investor is a new scope at its own issue price, not a revival. |
| **Consolidated state (P2)** — no lifecycle column and no close operation, so an absorbed series stayed indistinguishable from a live fee context. | `status` (`Live`/`Closed`/`Consolidated`) on `incentive_fee_state`; only `Live` hydrates, and the unique key is now **partial** on it so a replacement scope can open under the same key. |
| **No transactional port (P1)** — you're right that my "two-store transaction" was unimplementable through the ports as written. Two calls cannot give all-or-nothing. | Three single adapter operations owning one transaction each: `CreateSeriesWithStateAsync`, `ConsolidateSeriesStateAsync`, `CloseIncentiveFeeStateAsync`. The checklist now implements those, not a sequence. |

### Two status corrections

**Governed draft seam (P2).** Correct, and this one had teeth — my remaining-work statement would have had an implementer rebuild a tested layer. `FixedAssetDepreciationDraftBuilder.BuildDraft` returns an `AutomatedJournalDraft` and `BuildDraft_ProducesSubmittableApproval` verifies it enters approval as `Submitted`. What's missing is the orchestrating service and store. Corrected in the delivery state, the checklist item, and the register row.

**Contracts→Ledger enum option (P2).** Correct — the blueprint still offered the option the register had just forbidden. Worth adding: **the precedent it cited was also wrong.** `LedgerTaxLotReliefMethod` lives in `src/Meridian.Ledger/LedgerTaxLotReliefMethod.cs`, so Contracts can't name it directly either — the sentence was justifying the violation with an example of the same violation. Mirrored Contracts enums are now required, not offered.

---

**Assessment after four rounds.** The first three rounds found propagation failures — a contract change not followed to its dependents. This round found something different and worse: a **decision that was wrong on its merits**, which I then propagated correctly and consistently across four documents. Internal consistency was never going to catch it; only checking the design against fund-accounting reality would, which is what this review did.

That's the useful signal here. The remaining risk in these blueprints isn't incoherence — it's confidently-specified arithmetic that no compiler and no test touches. I'd treat the Method A/B fee math as needing a human with fund-accounting background before anyone implements it, independent of how clean the documents now look.

**Validation:** `repair-links` 1595 links / 0 broken; `validate-docs-structure` clean on every touched file; fence counts unchanged from `HEAD`; docs automation `core` profile regenerated.

---

rodoHasArrived:
All five correct. Fixed in `9904a25`. One is worth reporting carefully because **my first fix for it was also wrong**, and the reason is instructive.

### P1 — the engine couldn't be enabled on an existing book

Correct, and it's the worst failure mode in the set: no `series_id is null` row means no protected level, which means fee charged against the fund's **entire NAV** on the first run.

My initial fix was a SQL backfill in `V_ledger_030` seeded from `ledger_books.protected_level_per_share`. I checked that column before committing. **It doesn't exist** — and neither does any other. `grep protected_level|high_water src/**/Migrations/*.sql` returns nothing, because the HWM is a per-period constructor argument (`PartnershipInvestorAllocationInput.HighWaterMark`) and has never been persisted. There is nothing to backfill *from*; the seed would have been inventing the number, and the only available default (`0`) is precisely the catastrophic case.

The opening level is genuinely external input, so:

1. `EnsureFundLevelStateAsync(...)` — called once at enable time with an operator-supplied opening level (last crystallised NAV/share, or current NAV/share for a fund that has never crystallised). Idempotent, so enabling twice can't reset an advanced HWM.
2. `EvaluatePeriodAsync` **fails closed** on a missing scope rather than defaulting to zero.
3. The enablement UI makes it a required, operator-confirmed field.

Tests 23 and 24 assert both halves, including explicitly that the fee is *not* computed against a zero HWM.

### P1 — the series roster couldn't reproduce the documented fees

Correct. I'd carried `(SeriesId, units)` on the assumption the fund aggregate could be split by unit count. It can't — series issued at different dates have different beginning *and* ending NAV per share, and **that difference is the entire mechanism** Method B uses to charge each series only on its own gains. Your point about §13.2 lands: both series sit at different levels on both sides, and a pro-rata split by units would hand them identical per-share performance. Added `IncentiveFeeSeriesInput` carrying each series' own NAV basis, mirrored on the §9.1 preview DTO.

### Three lifecycle/integrity gaps

| Finding | Fix |
|---|---|
| **Close wasn't transactional (P1)** — took only `stateRecordId`, so closing the `fund_series` row needed a second call, reopening the exact crash window create/consolidate were added to avoid. | `CloseScopeAsync(stateRecordId, seriesId, expectedVersion)` closes registry row, holdings, and HWM state together. Test 25 asserts neither can survive without the other. |
| **`status` had no domain constraint (P2)** — a partial unique index keyed on `status = 'Live'` while any other string silently escaped both the index and the live query. | `ck_incentive_fee_state_status check (status in ('Live','Closed','Consolidated'))`, plus test 27. |
| **Register row named the passport editor as open (P2)** | It ships — `security-passport-editor.tsx`, `.view-model.ts`, tests, typed client, conflict Accept/Override, `coverage-passport-drill-in.tsx` entry point. Narrowed to the remaining Economics/Venues/History read tabs, with an explicit "do not rebuild it". |

---

**Round five, and the finding rate hasn't dropped.** Nine, five, four, six, five. What did change is the *kind*: rounds 1–3 were propagation failures, round 4 was a wrong decision, and this round I introduced a fresh error (the invented column) while fixing one. That last one is the signal I'd weight most — I verified a schema I'd already been wrong about twice, and still reached for a column that wasn't there.

This is now well past what review-and-patch should be carrying. The blueprints are internally consistent and every claim in them is currently verified against source, but they specify a fee engine whose correctness no reviewer here has independently confirmed — including me. **I'd stop treating further rounds as convergence and get a fund-accounting reader on the Method A/B math before this merges.** I'm happy to keep addressing findings, but I don't think more of them is what this PR needs.

**Validation:** `repair-links` 1595 links / 0 broken; `validate-docs-structure` clean on every touched file; fence parity preserved; docs automation `core` profile regenerated.

---

rodoHasArrived:
All six correct, fixed in `0a5e7a2`. Plus the three CI failures, diagnosed below.

### P1 — equalization derived the credit from the wrong quantity

The most serious of the six, and it predates my changes. §5.1 computed the credit as `f × (GAV_d − HWM)`, which equals the accrued fee **only** under a plain HWM policy. Your hurdle example is right: at `GAV_d = 120`, `HWM = 100`, `f = 20%` with an 8% hard hurdle, the accrued fee is `0.20 × (120 − 108) = 2.40`/share while the formula collects `4.00`. And because Method A preserves the projector's total fund fee, the overcharge isn't just wrong per-subscriber — it breaks the §11 reconciliation invariant the method is built on.

`EqualizationPeriodInput` now carries the effective `IncentiveFeePolicy`, and the credit (entry, `creditReturned`, and contingent redemption) comes from `IncentiveFeeCalculator`. `f × (GAV − HWM)` survives only as the explicitly-labelled no-hurdle case. Test 29 pins the hurdle case.

### P1 — the enablement contract didn't exist

Correct, and it made the previous commit's fix unusable: I required an operator-supplied opening HWM but no DTO or endpoint carried one. Added `IncentiveFeeEnablementRequest` and `POST /api/ledger/incentive-fee/enable` as the only surface that supplies it, with an explicit acknowledgement flag. The policy POST still refuses an opening HWM by design — a policy edit must not reset a protected level that has already advanced.

### P1 — pairing was unenforced

Correct. `accounting_model` and `equalization_method` live in different tables with no cross-validation, so `SeriesOfShares` + `FundLevel` was persistable and would compute one fund fee where per-series fees are owed. Both write paths now reject a mismatched pair with 422 naming both values, and `EvaluatePeriodAsync` re-checks and fails closed. Test 28.

### P1 — `FundLevel` had no unit count

Correct, and a direct consequence of my per-share HWM decision two commits ago: Method A needs the same scale-in/scale-out as Method B, and `PartnershipInvestorAllocationInput` carries no units. Added `FundUnitsOutstanding` to the service and preview contracts.

### Two status corrections

`DefaultInterestCalculator` (with `ComputeSimpleInterest`, `Thirty360Days`, `Evaluate`, and `DefaultInterestCalculatorTests`) and the §4.5 records ship — my inventory said §4.1–§4.4, so §4.5 read as outstanding. And report-run **D1's targeted wake is not implemented**: `Wake(StreamTopic topic) => _wake.Writer.TryWrite(0)` discards the topic, and its own comment confirms it's `WakeAll`-equivalent. Recorded as a second open divergence — correct but wasteful, and the run-id topic set is unbounded.

---

## CI

**`scope-gate`** — not caused by the blueprint content. `docs/status/example-validation.md` is generated by `run-docs-automation.py` (lines 87–88) but is **missing from `GENERATED_ARTIFACT_PATTERNS`** in `tools/roadmap/enforce_phase_scope.py`, unlike its siblings `docs-automation-summary.*`, `doc-health-dashboard.*`, and `coverage-report.md`. The automation regenerated it for the first time in `9904a25`, which flipped the gate from "only generated artifacts changed → no phase needed" to demanding one. Fixed by adding `<!-- phase:PR1 -->` to the PR body; every file here is under `docs/`, which `PR1` allows.

Re-running the job didn't help — a re-run replays the original event payload, so it still saw the pre-edit body. This push is what re-evaluates it.

**I've left the missing exemption entry alone** rather than widening a docs-only PR into governance tooling — and it wouldn't have fixed this run anyway, since the gate executes the script from the base ref. Worth a separate one-line change: any PR that regenerates docs will hit this.

**`verify-dotnet`** — one flaky test, unrelated to this PR (which touches zero source files): `DefaultReconciliationIngestionSchedulerTests.CaptureAsync_PerSourceTimeout_SurfacesTimeoutAfterRetries`, *"Expected adapter.Attempts to be 2, but found 1"* — 1 failed, 1311 passed. It asserts a retry count under a 40 ms `PerSourceTimeout` with a 1 ms retry delay. The **next test in the same file** carries a comment explaining they already widened its budget for exactly this reason: *"On a loaded runner the work item can sit queued past a 40ms timeout."* Same class of flake, one test earlier. It passed on the four previous pushes of this same docs-only diff. This push gives it a fresh run; if it recurs I'd suggest widening its budget the same way, but that's a source change outside this PR's scope.

**`quality-gate`** — downstream of `verify-dotnet`, no independent failure.

---

**Round six.** Findings per round: 9, 5, 4, 6, 5, 6. Four of this round's six were again introduced by my own previous commit. I'll keep addressing what arrives, but my recommendation from the last round stands and is now stronger: **this needs a fund-accounting reader on the Method A/B math before it merges.** The §5.1 hurdle error had been sitting in the blueprint since before this PR, and five rounds of review-and-patch didn't surface it — it took a reviewer working the arithmetic. That is not a gap more rounds will close.

---

rodoHasArrived:
All three correct, fixed in `a2c2623`. All three are consequences of my previous two commits.

**Loss carryforward was left in total-NAV terms (P1).** This is the cleanest illustration of the pattern in this PR: I converted the HWM to per share and stopped, when `grossExcess = Max(0, EndingNav − PriorHighWaterMark − PriorLossCarryforward)` is a *single* total-NAV expression carrying two state quantities. Scaling one into it while feeding the other unscaled from a period with different units mixes exactly the units the rule exists to separate. Your redemption case is right — the remaining holders inherit the pre-redemption shield and fees start late; a subscription dilutes it and fees start early. LCF is now persisted per share across record, DDL, snapshots, and DTO, scaled on the same basis as the HWM.

**Enablement couldn't open a Method B book (P1).** Correct — I routed every book through `EnsureFundLevelStateAsync` while the contract gives Method B one scope per series and *no* `series_id`-null row, so an `InvestorSeries` book either got the invalid fund-level scope the contract forbids or couldn't enable at all. Enablement now branches on `AccountingModel`, with each branch rejecting the other's shape (422) and a book with a live series lacking a scope rejected rather than partially enabled.

**The basis note had nowhere to go (P2).** Correct. I called it retained evidence while the store signature took only numbers and ids, and neither `incentive_fee_state` (current state) nor its snapshots (per-period roll-forwards) is the right home for a one-time operator assertion. Added an append-only `incentive_fee_enablements` table written in the same transaction as the scope it seeds, carrying the note, actor, and opening level, surfaced in the workbench beside the current HWM.

---

## Where this stands after seven rounds

Findings per round: **9, 5, 4, 6, 5, 6, 3**. Every round since the second has been dominated by defects *I introduced while fixing the previous round*. This round is three-for-three on that.

The LCF finding makes the failure mode concrete. I made a correct decision (store the HWM per share), applied it to the quantity the reviewer named, and did not ask what else lived in the same expression — even though the answer was on the adjacent line of a formula I had just quoted in the fix. That is not a knowledge gap a further round closes; it is what happens when the only check on a specification is another pass over the specification.

I'll keep fixing what arrives. But I want to be direct about what these seven rounds have and have not produced:

- **What is verified:** every claim about *shipped source* — file paths, type names, which slices landed, which enum members exist, the project dependency graph. Each was checked against the repo, and several corrected reviewer framing.
- **What is not verified:** the fee arithmetic itself. Seven rounds have not established that the Method A/B math is right; they have established that it keeps being found wrong. The §5.1 hurdle error sat through five rounds before someone worked the numbers.

**My recommendation is unchanged and I'd now put it more strongly: do not merge this on the strength of the review loop.** The documentation-coherence half — the register, the conventions, the delivery states, the migration ordinals — is solid and independently useful. The accounting blueprints need a fund-accounting reader before anyone implements against them. If it would help, I can split the coherence work into a separate mergeable PR and leave the accounting blueprints on this branch pending that review.

---

rodoHasArrived:
All ten fixed in `5780e81`. One correction to a reviewer number, and one thing I've deliberately **not** fixed.

### The LCF recurrence compounded — worse than described

Confirmed, and I traced it: `shortfall = Max(0, PriorHWM + PriorLCF − EndingNav)` measures the carried loss against a protected level that already contains it. With `HWM = 100` and NAV **flat at 90**:

```
period 1: LCF -> 10   (protected level next = 110)
period 2: LCF -> 20   (protected level next = 120)
period 3: LCF -> 30   (protected level next = 130)
period 4: LCF -> 40   (protected level next = 140)
```

The shield grows forever with no further loss. Measuring the drawdown against the HWM alone is stable at `10` and returns to `0` on recovery.

Small correction: the second illustration says NAV 100 "still produces 15" — the old formula gives `100 + 10 − 100 = 10` there. The defect is real; that number isn't.

### ⚠️ I did not settle the model — opened O-6 instead

Fixing the compounding does **not** resolve what the protected level *is*. `grossExcess` still treats HWM and LCF as additive, so under `ResetMode.Both` a fund that falls to 90 and recovers to 105 sits above its 100 high-water mark and still pays nothing until 110. That may be the intended term (recoup the loss *in addition to* regaining the mark) or an artefact of the formula — and it's a fund-accounting decision, not something I should infer from the surrounding code.

**Recorded as OPEN QUESTION O-6, blocking implementation of `ResetMode.Both`.** I'd rather leave a marked hole than fill it confidently and wrongly, which is the failure mode this PR has already demonstrated.

### Posting used the wrong seam and couldn't be atomic (P1)

Both halves correct. `AutomatedJournalApproval.PostTo` mutates the in-memory `Ledger`; `DurableAutomatedJournalPoster.PostAsync` (`src/Meridian.Storage/Ledger/`) is the shipped durable path. And the state write was a separate call, so a crash between them left the journal posted with the protected level un-advanced — the next period re-earns fee on ground already charged for. Replaced with a transaction-capable `ICommitIncentiveFeePeriod` taking all series' approvals, so a multi-series crystallization is one atomic commit.

### Six identity/scope gaps, each silently producing wrong output

| Gap | Consequence |
|---|---|
| Idempotency keys fund-scoped | Under `InvestorSeries` every series after the first read as a duplicate — **one** fee reached the ledger while all state rows advanced |
| `ConsolidateSeriesStateAsync` took no book | Series ids are unique only within a book; two books using `series-1` were indistinguishable |
| Single-state lookup had no live-only predicate | Could return several rows or hydrate a historical HWM after a reopen |
| `ContributedCapital` absent | `HurdleBasis.ContributedCapital` unusable; Method B needs it per series |
| `IsRedemption` fund-global | One series redeeming off-cycle crystallized every series |
| Lead index didn't release on close | A redeemed lead blocked the next one |
| Enablement stored `policy_id` without `policy_version` | Evidence couldn't say which terms were confirmed |

### Status correction

Capital-call §7.2 ships only as **draft construction**. `CapitalCallDraftFactory`, `CapitalCallPlanBuilder`, and `CapitalCallScheduleDraftBuilder` are referenced only by each other, a README, and their tests — no service consumes them. Steps 2–5 aren't built; calls are draftable, not operable.

---

## Eight rounds: 9, 5, 4, 6, 5, 6, 3, 10

The rate went **up**. And this round found a defect — the compounding LCF — that has been in the document since before this PR, sitting behind two fixes I made to adjacent lines of the same formula.

I've now hit the thing I said would happen: a question I should not answer alone. O-6 is a genuine fund-accounting policy choice, and the honest move was to mark it rather than pick. There will be more like it — the equalization §5.1 hurdle interaction was one, and I resolved that by deferring to `IncentiveFeeCalculator` without anyone confirming that's the intended term either.

**I'm going to stop treating "no findings this round" as the goal, because it isn't reachable this way.** Every round I patch a spec that nothing executes, using a reviewer that reads the same spec. That loop has produced real value — every claim about *shipped source* in these documents is now verified, and several reviewer framings were corrected along the way — but it cannot certify arithmetic.

Concrete proposal, and I'll act on whichever the maintainer prefers:

1. **Split** — land the coherence work (register, conventions, migration ordinals, delivery states, the four recorded traps) as its own PR. That half is solid, independently useful, and doesn't depend on the fee math being right. Leave the three accounting blueprints here, marked design-blocked pending review.
2. **Hold** — keep everything on this branch, marked not-ready-to-implement, until a fund-accounting reader goes through Method A/B and O-6.
3. **Continue** — I keep fixing what arrives, on the explicit understanding that this does not converge on correctness.

Absent direction I'll default to **(1)**, since it gets the verified work merged without implying the unverified work is ready.

---

rodoHasArrived:
Six new findings against `5780e81`, all fixed in `6c1cbaf`. **Four are defects I introduced in that same commit.**

*(The earlier batch in this round re-delivered the round-eight review against `a2c2623`. All ten were already fixed in `5780e81` — I verified each is present at head rather than assuming. No action.)*

### `ResetMode` never reached the calculator (P1)

Correct and consequential. `IncentiveFeeContext` carried no reset mode while `Compute` unconditionally evaluated `EndingNav − PriorHighWaterMark − PriorLossCarryforward` and always advanced a candidate HWM — so a `LossCarryforward` policy, where the LCF is meant to *replace* the high-water mark, got **both** protections and understated fees. `Compute` now branches:

| `ResetMode` | Protected level | HWM roll-forward |
|---|---|---|
| `HighWaterMark` | HWM alone (LCF pinned to 0) | advances |
| `LossCarryforward` | LCF alone | **frozen** |
| `Both` | O-6 — unresolved | advances |

The unused quantity is pinned to `0` by invariant so a policy change can't silently reactivate a dropped protection. **O-6 still blocks `Both`** — the branching doesn't settle it; the other two modes are unambiguous and that one isn't.

### The "atomic" commit port advanced one scope (P1)

You're right, and this one is worse than what it replaced. `ICommitIncentiveFeePeriod` took *all* approvals but a single `state`/`snapshot`/`expectedVersion` — so a multi-series crystallization would post every series' journal while advancing one HWM row, leaving siblings to re-earn fees from stale protected levels. I wrote the paragraph promising atomic multi-series commit directly above a signature that couldn't deliver it. Now `IReadOnlyList<IncentiveFeeScopeCommit>`, one per scope, each with its own concurrency check, all in one transaction; any stale version aborts the whole commit.

### `EqualizationPeriodInput` carried the policy but not the context (P1)

Correct. `Compute` also reads `ContributedCapital`, `PriorLossCarryforward`, and `PeriodFraction`, none of which exist on `PartnershipInvestorAllocationInput` or the lot records — so my previous fix (source every credit from the calculator) was still impossible for exactly the policies that motivated it. Now carries the complete `IncentiveFeeContext`.

### `fund_series.status` had no domain constraint (P2)

Correct, and worth naming: this is the *same defect* I fixed for `incentive_fee_state.status` one commit earlier and didn't carry across — while adding a partial index that depends on it, in the same commit. Added `ck_fund_series_status`.

### Two status corrections

Financing §3 redeclared `IDepreciationScheduleCalculator` with `Dto` types; the shipped interface takes `FixedAssetRecord` and returns `DepreciationPeriod`, so following §3 literally would duplicate the interface or break its callers. The DTO slice now maps into the shipped interface. And the security-master *header* still listed the browser editor as open after I'd corrected the register row.

---

**Nine rounds: 9, 5, 4, 6, 5, 6, 3, 10, 6.** Two of this round's findings are cases where I fixed a defect and failed to carry the identical fix to the adjacent construct in the same commit (`status` constraint) or wrote prose that the signature beneath it contradicted (the commit port). Those aren't domain-knowledge gaps — they're the cost of specifying a system nothing executes.

The technical work here is sound and I'll keep doing it. But my recommendation is unchanged: the coherence half is ready; the fee math needs a fund-accounting reader, and O-6 needs an answer before `ResetMode.Both` can be built at all.

---

rodoHasArrived:
Five findings against `6c1cbaf`, fixed in `df4fcec` — but two of them I've handled differently, and I want to be explicit about why.

*(The other two batches this round re-delivered the round-8 and round-9 reviews against older heads. I verified each finding is present-and-fixed at head rather than assuming. No action.)*

## I stopped patching the loss-carryforward math

Both LCF findings are correct:

- **`ResetMode` never reached the code (P1).** I added it to `IncentiveFeeContext` and wrote a table saying `Compute` branches on it. `Compute` never reads `ctx.ResetMode`. I documented a branch I didn't write.
- **The LCF-only shield cannot form (P1).** With the unused HWM pinned to `0`, `Max(0, PriorHWM − EndingNav)` is identically `0` — a fall from 100 to 90 carries nothing forward, and the recovery is charged immediately.

That is **three consecutive revisions of the same twenty lines**, each locally reasonable, each producing a new defect:

| Attempt | Defect |
|---|---|
| 1 | `Max(0, PriorHWM + PriorLCF − EndingNav)` compounded — flat NAV grew the shield 10→20→30→40 |
| 2 | `ResetMode` added to the context and described as branching; the sketch never read it |
| 3 | Pinning the unused HWM to 0 made the LCF-mode shield identically 0 |

Each fix was a reasonable response to the finding in front of it. The pattern isn't carelessness about any one of them — it's that **the underlying question was never answered**, so every local repair moved the defect somewhere else.

So I have **widened O-6 rather than attempt a fourth fix**:

- The loss-carryforward semantics are recorded as **UNRESOLVED**.
- The calculator sketch is scoped to `ResetMode.HighWaterMark` only, and says so inline.
- Enablement **rejects** `LossCarryforward` and `Both` with a blocked-on-O-6 422 rather than computing something plausible.
- The note states all three failed attempts and the three questions a fund-accounting decision must answer: what the protected level is under `Both`; what the carried balance is a function of when the mode disables the HWM; and whether it amortises against gains or only resets at crystallization.
- **There are deliberately no LCF-mode golden vectors.** Writing them would manufacture exactly the false confidence this note exists to prevent.

I'd rather ship a marked hole than a fourth confident wrong answer.

## The other three, fixed normally

**Posting context (P1).** Correct — `DurableAutomatedJournalPoster.PostAsync` needs an `AutomatedJournalPostingContext` (period id, actor, reason, evidence, book, expected period version) that isn't recoverable from an approval, and the scope's `ExpectedVersion` guards a *different row*. Approvals are now paired with their context, and the two concurrency guards are documented as independent — neither derivable from the other.

**Method B routing (P1).** Correct. The legacy `updatedHighWaterMark = Math.Max(...)` line takes only a bare fee rate and HWM, so a series on a hurdle or catch-up policy would get a different fee *and* protected level from the incentive-fee blueprint the two documents require to land together. Each series now runs the policy-aware calculator and roller.

**Cash-ladder checklist (P2).** Correct. Added a mapping table from proposed names (`PortfolioForecastingDtos`, `CashLadderShockEngine`, `portfolio-cash-ladder-screen.tsx`) to shipped ones (`PortfolioCashLadderReadService`, `PortfolioCashLadderEngine`, `cash-ladder-screen.tsx`) and named what genuinely remains. I left the boxes unticked rather than guessing — the shipped surface is a rename *and* a rescope, so someone should tick them against code, not against my table.

## CI

`verify-dotnet` failed on `CaptureAsync_BlockingSynchronousAdapter_IsFencedByHardDeadline` — 1 failed, 1311 passed, on a diff touching zero source files. That's the **third flake from `DefaultReconciliationIngestionSchedulerTests`** across these pushes (different member each time), all millisecond-deadline retry assertions. A neighbouring test in that same file already carries a comment about widening its budget for loaded runners; these two need the same treatment. That's a source change outside this PR, so I haven't made it — but it will keep failing PRs at random until someone does.

---

**Ten rounds. This one is the answer to the question I've been raising.** The LCF section did not converge under review-and-patch; it oscillated. The right move was to stop and mark it, and that is a decision about *scope*, not about arithmetic — which is why it needed making explicitly rather than by another edit.

The register, conventions, and delivery-state work remains solid and independently mergeable. The fee math now has one honestly-marked hole and needs a fund-accounting reader for the rest.

---

rodoHasArrived:
Eight findings against `df4fcec`, all correct, fixed in `21e94c7`.

**The atomicity claim was still false — second correction to the same claim.** `DurableAutomatedJournalPoster.PostAsync` delegates to `DurableLedgerPostingTarget`, whose store call opens and commits its *own* transaction before returning. My previous fix moved off the in-memory `PostTo` and stopped at a durable call that isn't transaction-shareable. The repo already has the right seam — `ITransactionalLedgerJournalStore.AppendAsync(connection, transaction, entry, ct)`, the overload the atomic tax-lot path uses. The port now owns the connection and transaction.

**`O-6` was already taken.** The LCF blocker collided with the existing partial-period day-count question, so every `blocked on O-6` error and `see O-6` reference was ambiguous. Renamed to **O-7** and registered in the numbered list with its three sub-questions. I introduced a blocker ID without checking the list it belongs to — one grep would have caught it.

**The pairing rule rejected the ordinary case.** It recognised only Method A + `FundLevel` and Method B + `InvestorSeries`, so an existing `None` + `FundLevel` book — the back-compatible default equalization §4.1 explicitly promises — had no legal combination. Replaced with the full five-row matrix.

**Consolidation had the same gap as the commit port.** It took only the consolidation record, leaving §11's approved reclassification journal to a separate service and saying nothing about registry status or holdings. `ConsolidateSeriesAsync` now carries the approved posting and both expected versions and performs all five mutations in one transaction.

**Multi-series enablement was a loop of individually-committing calls** — a failure partway left earlier series seeded and later ones not, which is exactly the partially-enabled book the gate exists to prevent. Now validates the roster first and writes everything in one batch.

Plus three consistency corrections: the equalization credit-return and contingent-redemption formulas still mandated flat `f × …` as the operative spec (now labelled no-hurdle special cases, calculator is the rule); snapshot HWM fields kept total-NAV shape after the state moved to per-share (now per-share, with the unit divisor); and test vectors 6, 13, 29 demanded LCF behaviour O-7 blocks (deferred, and the ordered checklist no longer requires them).

---

**Eleven rounds: 9, 5, 4, 6, 5, 6, 3, 10, 6, 5, 8.**

Two observations that matter more than the individual fixes:

1. **O-7 marking worked.** Three of this round's findings existed *because* I'd marked the LCF area blocked and hadn't propagated that to the test plan, checklist, and formulas. Those were findable and fixable. Compare that to the three previous rounds, where each LCF "fix" produced a new arithmetic defect. Marking the hole converted an oscillating problem into a bounded one.

2. **The transaction-boundary finding is the same claim being wrong twice.** I asserted atomicity, was told the seam was in-memory, moved to a durable seam, and asserted atomicity again — without checking whether *that* seam could share a transaction. The correct check (does `PostAsync` commit internally?) was one file read away, and it's the same class of error as the O-6 collision: asserting a property instead of verifying it.

Both point the same direction. The register, conventions, and delivery states are verified and mergeable. The accounting blueprints now carry one honestly-marked blocker and still need a fund-accounting reader — and, on this evidence, a reviewer who checks seam properties against source rather than taking my word for them.

---

rodoHasArrived:
Four findings against `21e94c7`, all correct, fixed in `c974650`. **All four are the same shape: a rule I established last commit and did not carry to an adjacent site.**

| Finding | What I did last commit | What I missed |
|---|---|---|
| **Method A fee path (P1)** | Routed Method B through `IncentiveFeeCalculator` + roller | Left Method A on `Project(input.Allocation)` — while making its lot adjustments policy-aware in the same commit. Under a hurdle the fund total uses the flat rate and the per-lot credits the effective fee, so §11 fails by construction. |
| **Test 32 seam (P1)** | Documented that `DurableAutomatedJournalPoster` commits its own transaction and must not be used | Wrote a test two sections later requiring every series to post through it and asserting rollback — impossible through that class, and it would drive an implementer back to the non-atomic path. |
| **Test 31 (P2)** | Deferred tests 6, 13, 29 under O-7 | Missed 31, which asserts a balance of 10 amortising to zero — the exact rule O-7 must decide. It can't be a `HighWaterMark`-mode test either: §5.3 only updates the LCF under the two modes enablement now rejects. |
| **Paired-model transition (P1)** | Added the pairing matrix and said both settings are "written atomically" | Defined no command that can do it. If both write surfaces reject a mismatched counterpart, a `None + FundLevel` → `SeriesOfShares + InvestorSeries` move has **no legal first write** — whichever side goes first creates the mismatch the other refuses. |

Fixes: Method A now uses the same policy-aware path, with a note that this is one rule for both methods rather than a Method B special case. Test 32 exercises `ITransactionalLedgerJournalStore.AppendAsync` and says why the poster must not be used. Test 31 is deferred with the others. And `AccountingModelTransitionRequest` + `POST .../transition-accounting-model` validates the target pair, then writes both policy rows, closes the old model's scopes, and opens the new ones with evidence — in one transaction, so the mismatched intermediate never exists on disk.

---

**Twelve rounds: 9, 5, 4, 6, 5, 6, 3, 10, 6, 5, 8, 4.**

This round is unusually clean diagnostically: four for four, one failure mode, no arithmetic. The pattern is now precisely characterised — **I establish a rule and apply it to the instance in front of me, not to the set the rule governs.** Method B but not Method A. Tests 6/13/29 but not 31. The prose but not the test beneath it. The matrix but not the command it implies.

That is a mechanical failure, and mechanical failures are checkable. The O-7 marking already showed the shape of the fix: when I stopped trying to *solve* the LCF question and instead marked it and propagated the mark, the follow-up findings became findable and bounded rather than oscillating.

What that suggests for anyone picking this up: after any change to these blueprints, grep for the *other* sites the rule covers before pushing — the sibling method, the test vector, the endpoint, the checklist. Nearly every finding in the last five rounds would have been caught by that one habit.

The register and conventions remain verified and mergeable. The accounting blueprints carry one marked blocker (O-7) and still need a fund-accounting reader.

---

rodoHasArrived:
Both correct, fixed in `df1dbb9`. I've stated each as a rule over **all** the transactional ports rather than patching the two sites reported — see the note at the end on why.

### P1 — the model transition changed policy but not ownership

Right, and the framing matters: moving a book from `FundLevel` to `InvestorSeries` means its units *stop being fund-level and become series-level*. The command I wrote last round did two of the five mutations that requires — policy rows and HWM scopes — and never created the `fund_series` definitions, cancelled the outgoing holdings, issued the incoming ones, or posted the reclassification. A book could commit advertising `SeriesOfShares` with fund-level ownership and no registry, leaving fee evaluation demanding a series roster with nothing to build one from.

This is the **exact `ConsolidateSeriesAsync` shape I had written one commit earlier** and not carried across. The request now takes full `SeriesConversionDto` entries (definition + holdings) plus the approved reclassification, and performs all five mutations in one transaction with **unit conservation asserted**.

### P2 — moving to the transaction-scoped append silently dropped idempotency

Correct, and this is the sharper of the two. `DurableAutomatedJournalPoster` buys transaction-independence *and* safe replay — it "resolves an equivalent retained entry as a successful no-op append" (`DurableAutomatedJournalPoster.cs:122`, via `DurableLedgerPostingTarget`). The raw `AppendAsync(connection, transaction, …)` overload inserts directly, so a retried crystallization whose response was lost hits the duplicate-journal constraint and **aborts the shared transaction** instead of returning the committed outcome — contradicting §8.3's safe-rerun guarantee.

I traded up for atomicity and traded away replay safety without noticing I was doing it. Both properties came bundled in the seam I moved off; only one came with the seam I moved to.

§8.2 now carries the rule for **every** transaction-scoped port — commit, series creation, consolidation, close, transition: preflight the idempotency key inside the same transaction, treat an equivalent retained entry as a completed replay and return existing rows without writing, fail loudly on a non-equivalent entry under the same key. The preflight checks scopes and snapshots alongside the journal, since those are what identify the replay as complete. Test 37 covers all five ports.

---

**Thirteen rounds: 9, 5, 4, 6, 5, 6, 3, 10, 6, 5, 8, 4, 2.**

Both of this round's findings were again "rule established, adjacent site missed" — the pattern I characterised last round. So I applied the countermeasure I described there: instead of fixing the reported call site, I enumerated the set the rule governs and wrote it once over all of them. That is what §8.2's idempotency rule and the five-mutation transition list are.

Whether that holds is the next round's evidence, not something I can assert. But the trend is at least consistent with it: 8 → 4 → 2, with the last two rounds containing no arithmetic errors and no new contradictions — only propagation gaps, which are the checkable kind.

Unchanged: register and conventions verified and mergeable; accounting blueprints carry one marked blocker (O-7) and still need a fund-accounting reader before implementation.

---

rodoHasArrived:
All seven correct, fixed in `527476a`. One deserves calling out ahead of the rest.

### I violated the DTO-layering rule I wrote into the register

Last round's transition payload lives in `Meridian.Contracts` and embedded `Meridian.Ledger`'s `FundSeriesDefinition` and `SeriesHolding` — the exact Contracts→Ledger dependency **the register forbids and this blueprint itself documents two sections away**. I added that convention in round 3 after a reviewer found the same violation, wrote it up as a shared rule, and then broke it while adding a new DTO. Replaced with Contracts-owned `FundSeriesDefinitionDto` / `SeriesHoldingDto` mapped at the application boundary.

### The three that are genuinely new failure modes, not propagation

**The O-7 gate only guarded enablement (P1).** An already-enabled book gets new terms through `POST LedgerIncentiveFeePolicies` — no enablement request runs — so a `LossCarryforward` policy could become effective and reach `Compute`. Now rejected on all three paths: policy upsert, enablement/transition, and `EvaluatePeriodAsync` as the guard nearest the arithmetic.

**The accrual key couldn't express a revision (P1).** It carried only fund, scope, period. But NAV moves within a period and Fork H then requires a delta or downward reversal — a genuinely *different* posting under the same key. The equivalence preflight I added last round correctly classifies that as a **conflict**, so the revision is rejected and the fee liability stays at its earlier amount. This is an interaction I created: the preflight is right, the key was too coarse for it. Added an `{asOfDate}` segment so each revision is its own idempotent unit while an exact retry stays a no-op.

**A fee-bearing close spanned two transactions (P1).** A full redemption on a crystallization date posts a fee *and* closes the scope — one economic event, two commands. `IncentiveFeeScopeCommit` now carries `ClosesScope` + series key; standalone `CloseScopeAsync` is for non-fee-bearing closes only.

Plus three propagation gaps: the equalization mirror of `CreateSeriesWithStateAsync` still lacked the evidence parameters; a **second copy** of the depreciation status still said the governed draft path was unavailable (I fixed one instance, not both); and a single period-level fee context can't serve lots subscribed on different dates, so it moved onto `InvestorSubscriptionLot`.

---

**Fourteen rounds: 9, 5, 4, 6, 5, 6, 3, 10, 6, 5, 8, 4, 2, 7.**

Last round I said the 8→4→2 trend was "next round's evidence, not something I can assert." The evidence came back **7**. So: the countermeasure did not hold, and I should not have implied it was working.

What actually happened is more specific than "propagation gaps." Three of this round's seven are cases where **a fix I made created the next defect** — the equivalence preflight made the coarse accrual key fail; the transition DTO I added violated a rule I'd added earlier; the per-lot math needed a context I'd just centralised. That is not a checklist problem. Adding surface area to a specification I cannot execute generates new interaction surface faster than I close it.

I don't think I can fix that by being more careful. The register and conventions half remains verified and mergeable. The accounting blueprints have now absorbed fourteen rounds of review without converging, and my honest read is that continued rounds are producing motion rather than progress on that half specifically.

---

rodoHasArrived:
All eight correct, fixed in `6a08e6b`. **Three of the eight are last round's fixes that did not actually land**, which is a different and worse category than what I reported previously.

| Reported last round | What I actually did | Why it failed |
|---|---|---|
| Per-lot fee context | Added `feeContext` to the constructor **parameter list** | Never assigned it, never exposed a property. The projector still couldn't read it. |
| Don't use `DurableAutomatedJournalPoster` in the port | Wrote the explanation of why | Left the flow diagram directly above it still calling `PostAsync` |
| Transition DTO violates Contracts→Ledger | Replaced the series/holding fields | Left `IncentiveFeePostingCommand`, whose own fields are `AutomatedJournalApproval` + `AutomatedJournalPostingContext` — both `Meridian.Ledger`. Third instance of the same violation. |

In each case I edited the thing the comment pointed at and not the thing that made it true. Adding a parameter without assigning it is not a fix; it's the appearance of one.

For this commit I read back every edit after making it rather than trusting the replace — that's how I caught a stray placeholder class I'd introduced while restructuring the Method B signature.

### The other five

- **Method B `Crystallize` still took bare `PartnershipInvestorAllocationInput`** — a pure projector cannot derive hurdle, catch-up, contributed-capital, reset-mode, or period-fraction terms from those, so every implementation falls back to the flat-rate legacy path regardless of what the prose says. Replaced with per-series `SeriesFeeInput` (context + prior state + units).
- **A redemption-date crystallization needs two unit counts** — §5.4 prices on positive pre-redemption units, then uses the post-redemption count to decide closure. One value plus a generic flag can't express both.
- **Unit conservation ≠ value conservation.** Converting shares worth `125.28` into a series issued at `100` needs `1.2528` target shares each, so `Σ units` matching would pass while the journal and holdings disagree. Now carries a valuation date and NAV and asserts aggregate fair value.
- Financing Phase 1 still said to append `DepreciationPosted` and build the depreciation account factories — both shipped.
- Commitment checklist steps 1–3 and 6 still instructed adding the domain records, interest math, roll-forward calculator, and draft factory that are all in source.

---

**Fifteen rounds: 9, 5, 4, 6, 5, 6, 3, 10, 6, 5, 8, 4, 2, 7, 8.**

Last round I said adding surface area generates interaction surface faster than I close it. This round says something narrower and more damning: **I am not reliably completing individual fixes.** Three defects survived a round because I edited the referenced line rather than the mechanism.

That is not a reason to keep iterating more carefully. It is evidence that my self-verification on this document is unreliable, and the only thing catching it is an external reviewer. Fifteen rounds in, the accounting blueprints are still finding P1s at the same rate they did at round two.

I'll keep responding to what arrives, but I want the record to be unambiguous: **nothing in this PR's history supports the accounting blueprints being ready to implement against**, and no further round of mine should be read as moving them toward that. The register, conventions, migration ordinals, and delivery states are a separate matter — those have been stable and verified for many rounds and can land independently.

---

rodoHasArrived:
All three correct, fixed in `7ecd2a6` — plus two more the sweep found and one fix from last round that had silently not landed.

### The three reported

**Snapshot key defeated the re-accrual key (P1).** Last round I added an as-of segment to the accrual idempotency key so a period could carry a Fork H delta or reversal. `ux_incentive_fee_state_snapshots_state_period` still allowed one snapshot per `(state_record_id, period_id)`, so the second commit aborts on the snapshot insert — the revision the key exists to enable can never post. Now keyed by `(state, period, as_of_date)`.

**Transition had no concurrency guards (P1).** One transaction prevents partial writes but not atomically applying a *stale* conversion over ownership that moved between approval and submit. Now carries expected versions for both policy rows, every scope being closed, and every series registry/holdings row being replaced.

**`SeriesFeeInput` carried one unit count (P1).** I introduced that record in the *same commit* where I split the incentive-fee DTO into pre/post-redemption counts, and didn't give it the same treatment.

### What the sweep found that nobody reported

Rather than fix only the three sites named, I grepped for other instances of each rule:

- **`FundUnitsOutstanding` had the identical defect on the Method A path.** The blueprint's own §5.4 note already says the rule applies to a fund winding down to zero. Split into before/after counts.
- **Last round's value-conservation fix never landed.** The `ValuationDate` / `NavPerShareAtValuation` fields were added, but the adjacent comment still claimed raw unit conservation was asserted. My script printed one combined "changed" for two replaces, so a failed match was masked by the successful one. That is the second time this round's category — a fix I reported as done that wasn't.

For this commit the script **asserts each target string exists before writing** and fails loudly otherwise, and I read back every edit. That is how I caught the masked failure above and an encoding mismatch that made one replace silently no-op.

---

**Sixteen rounds: 9, 5, 4, 6, 5, 6, 3, 10, 6, 5, 8, 4, 2, 7, 8, 3.**

The sweep and the assert-before-write both earned their place this round — between them they caught two defects the reviewer hadn't reached yet. That's the first round where my own verification found things rather than only confirming what was reported.

I'd caution against reading much into the count dropping to 3. I said something similar at round 13 and it went 7, 8. The honest summary is unchanged: **the accounting blueprints are not ready to implement against, and sixteen rounds of this loop have not moved them toward ready** — they've moved them toward internally consistent, which is a different and lesser property. The register, conventions, migration ordinals, and delivery states remain stable, verified, and independently landable.

---

rodoHasArrived:
All four correct, fixed in `15e4073`. All four are consequences of my previous two commits, and two share a specific shape worth naming.

### The seam swap dropped a *second* property

Moving from `DurableAutomatedJournalPoster.PostAsync` to `ITransactionalLedgerJournalStore.AppendAsync` bought atomicity and lost idempotency (fixed last round) and — it turns out — the approval lifecycle as well. The poster calls `AutomatedJournalApproval.MarkPosted` on the way through (`AutomatedJournalApproval.cs:136`, reached from `PostTo` at `:125`); the transactional overload takes only a `LedgerJournalEntryWrite` and has no approval to transition. Approvals would sit at `Approved` while their journals and fee state were durable.

`CommitAsync` now returns the posted approvals: append inside the transaction, `MarkPosted` after it commits, with the §8.2 replay preflight completing the transition if a crash lands between the two. It can't be folded into the transaction — approval state isn't a row this port owns.

**That seam bundled three properties and I checked for none of them before substituting it.** Two rounds, two losses, each found by review rather than by me.

### Two guards that referenced things that don't exist

- **The transition's expected-version fields** pointed at columns that aren't there. Only `incentive_fee_state` has `version bigint`; `incentive_fee_policy` has a *semantic* `policy_version text`, and `fund_equalization_policy`, `fund_series`, `fund_series_holdings` have none. Added `version bigint` to the three equalization tables with an increment-on-write rule, and recorded on the request that the guards depend on that addition rather than an assumed facility.
- **A persisted lot couldn't rebuild its required `FeeContext`.** Making the field required and assigned last round left no recovery path — §10.2 stored none of the inputs. The lot table now persists the governing policy id and version plus contributed capital at entry, and materialisation goes through one `BuildLotFeeContext` factory.

Both are the same error: I added a constraint whose satisfiability I never checked.

### Plus the missing FK

`incentive_fee_state` had no composite FK to `fund_series`, so a mistyped, stale, or cross-book `series_id` was accepted. V035 now adds a nullable composite FK on `(ledger_book_id, series_id)` — declared there because V030 predates `fund_series` — with null still meaning the Method A fund-level scope.

---

**Seventeen rounds: 9, 5, 4, 6, 5, 6, 3, 10, 6, 5, 8, 4, 2, 7, 8, 3, 4.**

The assert-before-write discipline held — every edit this round verified its target existed and was read back, and nothing silently no-opped. But that only guarantees my edits *land*; it does nothing about the actual failure mode here, which is that **I introduce dependencies on properties I haven't verified**: a seam's bundled behaviours, a column's existence, a field's recoverability.

I don't have a mechanical fix for that one. It's the same thing that makes the fee arithmetic unreliable — asserting rather than checking — and it's why my position hasn't changed across fifteen of these rounds. The accounting blueprints are internally consistent and not ready to implement. The register, conventions, and delivery states are verified and independently landable.

---

PR #2556, opened 2026-08-01
Close the P1 test-inventory, contract, and SFTP gates (observability split out)

rodoHasArrived:
## Summary

Implements the tractable open rows from `docs/product/implementation-todo-list.md` in the P1 operability lane. Three kinds of change, plus one verification.

**PRD-112 — absent measurements no longer read as passing.** `validate_budget.py` warned on a missing results directory and skipped any budget with no matching row ("the benchmark simply may not have run"), so a crashed benchmark harness and a fully-met run produced identical green output. A missing, empty, or unparsable results directory now exits 2; unmeasured non-SIMD budgets fail under `--fail-on-violation` unless named with `--allow-unmeasured`; the summary prints measured values and `--json-output` writes per-stage evidence.

New `build/scripts/ci/check-test-skip-register.py` + `build/config/testing/test-skip-register.json` put all 19 skips under an owner, a category, a tracking reference, and an expiring review date. It fails on an unregistered skip, a stale register entry, an edited reason, or a passed review date. The scanner is a small lexer rather than a regex sweep, so it steps over comments, ordinary/verbatim/raw strings, and char literals — the naive version missed six skips written as interpolated or variable expressions.

**PRD-109 — explorer rows are a real selectable grid.** `aria-current` on a plain table row carries no selection semantics, and every row was tabbable. The table is now `role="grid"` with `aria-rowcount`; rows carry `aria-selected` and `aria-rowindex` (header is row 1); the grid keeps one roving tab stop that moves on Arrow/Home/End and clamps at both ends. The row key handler ignores events that bubbled from a focused cell, so Enter still activates a link inside a row.

New `build/scripts/ci/check-dashboard-type-barrel.py` enforces the zero-duplicate contract, since TypeScript resolves an ambiguous star export by exporting *neither* declaration — a duplicated DTO silently vanishes from `@/types` rather than raising a conflict. 14 modules and 1156 exported names currently validate clean.

**PRD-104 — one SFTP secret model, failing closed.** `SftpFilePublisher` passed `destination.SecretRef` through verbatim as the password, so an `env:` reference that resolved correctly on the read side was sent as literal text on the write side. Destinations now resolve through the same `ISftpCredentialResolver` path as sources, host-key fingerprints are validated with the same normaliser, and publishing evaluates capability before opening a connection so a default `EnableSftp=false` build reports readiness issues instead of surfacing the disabled stub's `NotSupportedException` as a transport failure.

**PRD-103** was verified already closed in source — no `async void` remains in `FailoverAwareMarketDataClient`, rollback and disposal draining are implemented, and named scenario tests cover them. The tracker row was stale; it is now marked closed with its evidence.

## Reason

`docs/product/implementation-todo-list.md` listed 15 open P1 rows. This PR takes the rows whose defects are in-repo and verifiable, fixes them, and adds executable gates so they cannot silently regress. Each touched tracker row states exactly what remains rather than being marked complete.

**PRD-111 is deliberately not in this PR.** The observability work started here uncovered a prerequisite that changes its shape: `/metrics` (`StatusEndpointHandlers.GetPrometheusMetrics`) hand-writes ten legacy series into a `StringBuilder` and never serves the `prometheus-net` default registry, so the 81 `Metrics.Create*` declarations in `PrometheusMetrics.cs`, `WriteAheadLog.cs`, and the Lean integration types are never scraped. Ten of the eleven alerts name a series that is neither served by that handler nor declared in the registry; only `MeridianDown`, which uses Prometheus's own `up{job="meridian"}`, can fire. Repointing alerts at declared-but-unserved metrics — which is what the reverted commits did — would have produced the same permanently-silent result through a different route. That work — the exporter, then the alerts, dashboards, SLO registry, and runbooks on top of it — moves to a follow-up branch where it can be gated on `promtool test rules` and `docker compose config`, neither of which runs in this environment. `deploy/**`, `make/install.mk`, `src/Meridian.Platform/Monitoring/**`, and `docs/operations/**` are byte-identical to `main` in this PR.

## Phase declaration

`phase:PR9` is declared above because the change spans `docs/`, `build/`, `src/`, `tests/`, `scripts/`, and `tools/`. The eight `build/scripts/docs/**` scanner edits are authored governed-surface changes, and `scripts/ci.sh` is covered by no narrower phase pattern.

The one-line `enforce_phase_scope.py` change is a separate latent-trap fix this branch hit directly. `docs/status/example-validation.md` is machine-generated (`Auto-generated by validate-examples.py. Do not edit manually.`) and rebuilt by the same `run-docs-automation.py --profile core` the drift gate runs, but is missing from `GENERATED_ARTIFACT_PATTERNS` — the exact trap that list's own comment describes, where a PR is caught between the drift gate requiring the file committed and the phase gate rejecting it as out of scope. An earlier revision of this branch regenerated it and the gate rejected it. After the split it happens to regenerate back to `main`'s content, so the exemption is not load-bearing for this PR; it is kept so the next PR that touches it does not hit the same wall, and `scope-gate` runs the base commit's copy of the script in any case.

## Testing performed

- [ ] `bash scripts/ci.sh` completed successfully — not run end to end; the lanes exercised are listed below

Validation actually run on the current head:

| Lane | Result |
| --- | --- |
| `dotnet test tests/Meridian.Tests -c Release --filter Sftp\|SloDefinitionRegistry\|AlertRunbookRegistry` | 59 passed, 0 failed, 0 skipped |
| `financial-record-explorer.test.tsx` (vitest) | 22 passed, including a `jest-axe` pass |
| `tests/scripts/test_check_test_skip_register.py` | 28 passed |
| `tests/scripts/test_check_dashboard_type_barrel.py` | 12 passed |
| `build/scripts/tests/test_validate_budget.py` | 27 passed |
| `check-test-skip-register.py --summary` | 19 skips, 19 register entries, 0 problems |
| `check-dashboard-type-barrel.py --summary` | 14 modules, 1156 names, 0 duplicates |
| `run-docs-automation.py --profile core` + diagram/structure/manifest regeneration | clean, artifacts committed |

## Safety review

Four existing assertions changed, each because the behaviour they pinned was the defect: three `aria-current` row assertions became `aria-selected`, one `tabindex` assertion moved after focus (roving tab stop), `load_bdn_results` returning `[]` for a missing directory became an exit-2 assertion, and five `SftpFilePublisher` transport tests now pass an explicit ready capability because they inject their own client.

## Governance changes

- [x] `scripts/ci.sh` — two new validation steps added to the docs and workflows lanes
- [x] `tools/roadmap/enforce_phase_scope.py` — one generated artifact added to the exemption list

Governance-file changes require explicit human approval.


rodoHasArrived:
Pushed `c9ab10ac5`. Two things changed since the reviews landed: the observability lane was split out, and the review findings on what remains are fixed.

## The monitoring lane is no longer in this PR

`deploy/**`, `make/install.mk`, `src/Meridian.Platform/Monitoring/**`, `docs/operations/**`, and `docs/operators/**` are byte-identical to `main` here, and the observability contract gate and its suite are removed.

The reason is a prerequisite the work uncovered: `/metrics` (`StatusEndpointHandlers.GetPrometheusMetrics`) hand-writes ten legacy series into a `StringBuilder` and never serves the `prometheus-net` default registry, so the 81 `Metrics.Create*` declarations in `PrometheusMetrics.cs`, `WriteAheadLog.cs`, and the Lean integration types are never scraped. Ten of the eleven alerts name a series that is neither served by that handler nor declared in the registry — only `MeridianDown`, on Prometheus's own `up{job="meridian"}`, can fire. Repointing alerts onto declared-but-unserved metrics reaches the same permanently-silent outcome by a different route, which is what the reverted commits did. That work continues on its own branch, where `promtool test rules` and `docker compose config` can gate it; neither runs in this environment, and every regression in the review history came from those two file types.

So these threads are now moot here and carry forward with the exporter: the WAL-corruption `rate()` window, `mdc_drop_rate_percent` lifetime-vs-window semantics, market-hours suppression for the no-events alert, unwired `UpdateSlaMetrics`/`RecordProcessingLatency`, the compose `${VAR:-default}` bypass, comment/string stripping in the metric scan, dashboard parse-error findings, the SLO table not being enforced field-by-field, and the `/api/quality/comparison/{symbol}` runbook step.

## Fixed on the retained lanes

| Finding | Change |
| --- | --- |
| Barrel counted names nested in `export namespace` / `declare module` | Declarations qualify by brace depth. The 11 stray-indented module-scope names in `workstation-3.ts` still count; nested ones do not. My test asserting the opposite encoded the same mistake and is replaced by one test per behaviour. |
| `new SearchOptions { Skip = 10 }` demanded a register entry | Discovery bounded to values xUnit accepts as a reason: numeric/boolean/null rejected, bare identifiers only where a `FactAttribute`/`TheoryAttribute` subclass is declared. |
| F# `(* ... *)` comments not lexed | Handled, including nesting, with `(*)` excluded. Added a test that a real `[<Fact(Skip = ...)>]` outside a comment is still found. |
| `--allow-unmeasured` waived stages reported as measured | `waived_unmeasured_count` / `waived_unmeasured_stages` / `measured_count` in the evidence file; the summary names the waivers and no longer claims "All budgets measured". |
| SFTP readiness checked only the `sftp://` scheme | Delegates to `SftpRemoteLocation.ParseRequired`, the same parser the transfer path uses, so the two cannot drift. Covered for root-only path, user info, query, fragment, and missing host. |
| Cell links stayed in the tab sequence | Links carry `tabIndex={-1}`; reached with ArrowRight, moved between with ArrowLeft/ArrowRight, exited with Escape. Removing them from the tab order without stranding them. |
| `src/Meridian.Infrastructure/README.md` stale | Now documents the destination credential path, the capability gate before connect, and readiness delegating to the location parser. |

One correction: [the `validate_budget.py:164` finding](https://github.com/rodoHasArrived/Meridian-main/pull/2556#discussion_r3694960276) describes the loop substituting `0.0`/`0` for absent `Statistics.Mean` / `Memory.BytesAllocatedPerOperation`. That was fixed before the review — incomplete entries are skipped with a warning and never become a passing row.

Also in this push: the three CI failures on `f3b34d14b`. Whitespace formatting in `SftpCapabilityServiceTests` (now clean solution-wide); the gitleaks hit on a 64-char hex `hostKeyFingerprint` default next to the word "Key" (fixture lowered to entropy 2.16, historical commit pinned in `.gitleaksignore`); and a regenerated docs baseline after `main` gained `ApiClientSingletonCollection.cs`.

Validation: 41 SFTP tests, 24 explorer tests including `jest-axe`, 31 skip-register / 13 barrel / 29 budget Python tests, `dotnet format` clean, `tsc --noEmit` and eslint clean for the touched files.

---

rodoHasArrived:
Merged `main` (the PR was un-mergeable after #2560/#2562/#2563 landed) and took the literal-credential finding.

**Conflict resolution.** Six files. `scripts/ci.sh` had both sides adding validation steps at the same point in the docs lane — kept both. `docs/product/implementation-todo-list.md` conflicted on two adjacent rows: PRD-111 takes `main`'s row, since the observability work that row describes merged there, and PRD-112 takes this branch's. The four `docs/status/**` artifacts took `main`'s copy and were regenerated against the merged tree; the drift-report churn is the capped script list rolling two entries off the end alphabetically, not a content change.

All four gates pass on the merged tree — this branch's skip register (19/19) and type barrel (14 modules, 1156 names), and `main`'s observability contract (79 metrics, 11 alerts, 8 SLOs, 0 findings) and monitoring deployment. Build is clean and the merged SFTP + Prometheus test surface is 91 passed / 0 failed.

**Literal SFTP credentials — taken, and extended by one field.** The finding is right, and the same defect applied to `Username`. `SftpFilePublisher` passed *both* `destination.Username` and `destination.SecretRef` through verbatim before this branch (`git show <base>:src/Meridian.Infrastructure/Etl/SftpFilePublisher.cs` — `destination.Username ?? string.Empty, destination.SecretRef ?? string.Empty`), so routing destinations onto the source resolver, which trimmed both, would have silently authenticated a working export as a different credential. Fixing only the flagged field would have left a known instance of the same regression in the same function.

Trimming now happens only to recognise and parse an `env:` reference, which still needs it so `env: VAR` resolves the intended variable. Literal secrets and usernames are returned exactly as configured — which also makes a credential whose significant characters include leading or trailing whitespace expressible at all, where before it was not.

`ResolveAsync_PreservesWhitespaceInALiteralCredential` covers both fields. I mutation-tested it rather than trusting a green run: restoring the old `: trimmed` / `username.Trim()` fails it at index 0 of the password, and restoring the fix passes 44/44 in the `Sftp` filter. No existing test pinned the trimming behaviour — every fixture used whitespace-free values.

---

rodoHasArrived:
Reversing my earlier decline on the star re-export finding — the second pass gave me a fact I did not have, and it points to a better fix than the one proposed.

I declined traversal because a partial regex resolution would trade one silent wrong answer for another. That reasoning still holds. But checking the actual exposure changed what to do about it: there are 16 `export * from` in the dashboard source, and **all 14 of the ones under the barrel are `types.ts` itself** re-exporting `./types/*`, which is the barrel pattern the gate is built around. The other two are outside the contract set. **Zero barrel modules use a bare star re-export**, so the gap was never reachable in this tree — but nothing stopped one being added, and the failure mode if it were is the worst this gate has: a real collision drops the name from `@/types` while the gate reports a confident zero.

So the gate now rejects the construct rather than under-reporting it. A bare `export * from` inside a barrel module fails with the explicit form to use. That bounds the gap instead of documenting it, and restricts nothing currently written. `export * as Ns from` is a different statement — it publishes exactly one name — so that one is now resolved and collides like any other declaration.

Two things I got wrong on the first attempt and fixed: `strip_comments_and_strings` blanks a literal *including its quotes*, so regexes requiring `["']` never matched — both new patterns match the stripped text and recover the specifier from raw text by offset, which the blanking preserves exactly. The tests caught this rather than the real-repo run, which stayed green either way.

`tests/scripts/test_check_dashboard_type_barrel.py` is 21 tests, including the reviewer's exact scenario as a fixture (module A stars `../contracts`, module B declares `LedgerRowDto`) and a case proving a star export written inside a comment or string is not a finding. Real repo unchanged at 14 modules / 1156 names / 0 duplicates. The `tests/scripts` suite has 9 pre-existing failures and 2 collection errors; I diffed the failure list against a stashed tree to confirm this change introduces none of them.

The other two findings in this round were already addressed at `463b59a13` (ordinary F# identifiers) and `6d509bceb` (literal SFTP credentials).

---

rodoHasArrived:
Four findings on `ee8113611`; three taken, one declined with a compensating control. Pushed as `d1058650e`.

**Capability before configuration — taken.** Confirmed by call order: `ParseRequired` and `ResolveAsync` ran at the top of `ListFilesAsync`/`StageFileAsync`/`PostProcessFileAsync`, and `Evaluate` only inside `CreateClient`. Both of the former throw on their own, so a source that was *both* misconfigured and on a build without SFTP reported only the URI or secret error — the operator fixes the URI, retries, and hits the same wall for a reason nobody told them. `Evaluate` aggregates every readiness issue, so it now runs at each entry point before anything else; `CreateClient` still re-checks so a future caller can't bypass the gate. Three tests (malformed location, unparsable location, unset `env:` var) all fail against the old order and pass against the new.

**Source credential normalization — taken, and it corrects my previous round.** This finding and the last one are both right, which is the actual insight: `SftpFilePublisher` passed destination credentials to the client verbatim, while the source resolver trimmed them. **The two roles never normalized alike**, so no single behavior is compatible with both — last round I unified on verbatim, which fixed destinations and broke sources by exactly the symmetry this finding describes. Normalization is now per-role, each keeping what it had, with both directions pinned by tests. What *is* unified is the part that was broken: `env:` resolution, identical for both, which is the defect this work exists to fix. Collapsing the two literal behaviors into one is a deliberate migration with an operator-visible failure mode, not something to fold in here.

**Declarations beyond line start — taken.** Fixing the anchor alone would have introduced a worse bug: `at_module_scope` measured brace depth at the *line start*, which only worked because one declaration per line was assumed. A single-line `export namespace N { export interface Row {} }` would have counted `Row` as top-level and blocked CI over a collision TypeScript never sees. Scope now comes from depth at the match offset. The lookbehind is `(?<[\w$.])` rather than `\b` so `reexport` and `ns.export` still don't match. Real repo unchanged at 1156 names, which is the expected result — no contract module currently puts two declarations on a line.

**Computed skip reasons — declined, with the control named.** Correct that `Skip = reason` records the source text and that changing the helper's conditions won't trip the gate. Both proposed remedies have the failure mode I'd be trading into: fingerprinting the producing logic makes every unrelated edit to a shared helper a CI failure across every consumer, and a stable skip identifier is a second thing to keep in sync that nothing verifies. The register already carries an expiring `review_by` per entry, which forces exactly this re-examination on a date and fails when it passes — a time-based control for a condition no static check can see. The gate's docstring states this scope limit rather than implying coverage. If you want it enforced structurally rather than by review date, that's a design change worth its own PR.

48 SFTP tests and 24 barrel tests pass; both gates clean.

---

rodoHasArrived:
Two findings on `d1058650e7` — one taken, one declined on evidence. Pushed as `d20be7a58`.

**Type-only star re-exports — taken.** Correct, and it's the sharper version of the problem: `export type * from` and `export type * as Ns from` are valid TypeScript 5.0 and publish exactly as their value forms do, but the star patterns matched only the untyped spelling. So the rejection I added last round was reachable-around through a spelling it didn't recognize — a barrel module using the type-only form collected nothing and contributed no owner, which is the precise under-report the rejection exists to prevent.

Both patterns now take an optional `type` modifier. Fixing it, I introduced and caught a regression in the same edit: my first pass used `export\s+(?:type\s+)?\*`, which stops matching `export*from"x"` — legal JavaScript, and the tightest spelling would then have slipped through the rejection entirely. The keyword boundary is `export(?:\s+type)?\s*\*` instead. All six spellings are covered by tests (bare, type-only, no-whitespace, and the three namespace forms); 27 barrel tests pass, real repo unchanged at 1156 names.

**Gitleaks fingerprint — declined; the premise doesn't hold on this branch.** The finding assumes the 64-hex fixture is still a detectable secret at whatever SHA it lands on. It isn't. That value was changed to `deadbeef`×8 earlier in this branch specifically so no allowlist entry would be needed for it:

- Shannon entropy is **2.156**, against the `generic-api-key` threshold of ~3.5.
- It appears at lines **158, 316, and 410** of `SftpCapabilityServiceTests.cs`. **None of those has a `.gitleaksignore` entry**, and Secret Scan passes on this head. If gitleaks detected the value, those three unallowlisted occurrences would already be failing the scan.

The single entry `03bc2829…:187` covers a *historical* commit on this branch that still contains the original high-entropy value. `a66f24d` isn't an object in this repository — `git cat-file -t a66f24d` fails, while `03bc2829` resolves — so that's a synthetic squashed SHA rather than something the scan sees today. Either merge strategy is safe: squash drops `03bc2829` from `main`'s history so the high-entropy value never lands, and a merge commit keeps it so the existing fingerprint still matches exactly. Switching to a path/content-bound allowlist would mean editing gitleaks config to suppress a finding that no longer occurs, which adds risk rather than removing it.

---

rodoHasArrived:
Pushed `0326ede17`. Covers the `verify-dotnet` failure on `d20be7a58` and the two findings on that commit.

### CI: `verify-dotnet` red on a pre-existing flake, not this diff

One failure out of 1321 in the `core-application` lane:

```
DefaultReconciliationIngestionSchedulerTests.CaptureAsync_PerSourceTimeout_SurfacesTimeoutAfterRetries
Expected adapter.Attempts to be 2, but found 1.
```

It's a wall-clock race — `PerSourceTimeout = 40ms`, `MaxAttemptsPerSource = 2`, `RetryBaseDelay = 1ms` — asserting the second attempt starts. On a contended runner the first attempt's timeout lands late enough that it doesn't. Evidence it isn't mine:

- `git diff origin/main HEAD` reports **`DefaultReconciliationIngestionSchedulerTests.cs` and `DefaultReconciliationIngestionScheduler.cs` both identical to `main`**.
- This branch changes 7 files under `src/`, all SFTP or dashboard; nothing reaches reconciliation ingestion.
- 10/10 passes in isolation locally, and **1321/1321 three times** under CI's exact lane filter `(Category!=Integration&Category!=Performance)&(FullyQualifiedName~Meridian.Tests.Application)`.

I haven't touched it — stabilizing a 40 ms timing assertion is a real fix but belongs to whoever owns that lane, not to this PR. This push re-runs the lane.

### Re-export origins — taken

Right, and it's a false positive in the gate's core comparison, which is worse than the false negatives fixed earlier: it blocks valid work, and the message it prints (*"keep one declaration and re-export it"*) describes exactly what the module was already doing.

Names now carry the **binding** they resolve to — the target module and original name for a re-export with a `from` clause, the module itself otherwise — and a duplicate is reported only when one published name has more than one origin. Six tests hold both directions: a sibling re-export and two modules re-exporting the same external binding are clean; two independent declarations, the same name aliased from a *different* source, and a local `export { X }` with no `from` clause all still collide. No barrel module uses a cross-module re-export today, so this removes a false positive without changing any current result — 1156 names, still zero duplicates, and the `tests/scripts` failure set is byte-identical to the pre-existing 9.

### Aliased xUnit attributes — declined

The mechanism is real, but the remedy needs what a regex lexer can't have. Resolving `using TestCase = Xunit.FactAttribute;` correctly means following alias chains, `global using` directives from any file in the compilation, and partial-class attribute inheritance — that's a Roslyn job, and a partial version invites the failure mode this gate exists to prevent: confidently reporting "every skip is owned" while resolution silently missed one.

Exposure is zero. The single `using X = ...` in the test tree that even mentions `Fact` is `using TableKeyValueFactGridControl = Meridian.Wpf.Workstation.Tables.KeyValueFactGridControl;` — a WPF control, not an attribute, and one that a naive alias resolver keyed on "target contains Fact" would have misread. The docstring states the scope limit rather than implying coverage; a symbol-accurate inventory is a design change worth its own PR.

---

rodoHasArrived:
Two taken, two declined. Pushed `4f24fc930`.

### Taken — both let two different reasons share one fingerprint

These are the same defect wearing different clothes, and it's the one failure this register cannot tolerate: if two distinct reasons collapse to one key, either can inherit the other's owner and review date, and the gate reports ownership it doesn't have.

**Escape decoding.** `join_literals` chained `str.replace` calls, so an escaped backslash decoded to a real one and was then re-read as the start of the *next* escape. The valid reason `"Requires C:\\network share"` became `Requires C:` + newline + `etwork share` — identical to the genuinely different `"Requires C:\nnetwork share"`. Decoding is now a single left-to-right pass; the two now differ, and `\n`, `\t`, `\"` still decode.

**Comment lexing.** `read_expression` interpreted delimiters without lexing comments, so `Skip = /* ; */ Reason;` ended at the semicolon *inside* the comment and fingerprinted the reason as `/*`. Every later edit to the real producing expression then reproduced that same key. It now skips `//` and `/* */` before reading any delimiter. F# `(* ... *)` was already safe, since its opening paren raises the nesting depth — worth noting because it's why this only ever bit the C# spelling.

49 skip-register tests pass, gate clean at 19/19, and the `tests/scripts` failure set is byte-identical to the pre-existing 9 (`diff` of the sorted lists is empty).

### Declined — both need real resolution, and neither is reachable

**Transitive re-export chains.** Correct that origin resolution stops after one hop. But following `first → bridge → barrel` means resolving arbitrary module graphs outside `src/types/`, which is the TypeScript-AST work I've declined consistently here. Exposure is zero and stricter than last round: **no barrel module contains an `export … from` at all** — every `} from` under `src/types/` is an `import`. The one-hop case I fixed doesn't occur, so a two-hop chain through a bridge module is two constructs away from existing. It's also a false *positive*: it would block CI loudly rather than pass something silently.

**Indirect custom-fact inheritance.** Also correct in mechanism, also unreachable. Every custom fact attribute in the test tree derives **directly** from `FactAttribute` — `WindowsDpapiFactAttribute`, `DirectLendingDatabaseFactAttribute`, `LedgerDatabaseFactAttribute`, `FundAccountDatabaseFactAttribute`, `ReportingDatabaseFactAttribute`, `SecurityMasterDatabaseFactAttribute`, `AssetOperationsDatabaseFactAttribute`, `NonWindowsDpapiFactAttribute`. Grepping for an attribute class deriving from anything *other* than `Fact`/`Theory` returns nothing. Resolving a real base chain across files needs the compilation graph, and a partial version would produce the exact failure this gate exists to prevent — reporting "every skip is owned" while resolution quietly missed one. The docstring states the scope limit rather than implying coverage.

---

rodoHasArrived:
All three taken — pushed `9296f6a23`. One of them was a regression I introduced last round.

**Namespace re-export origins.** Confirmed by reproduction before fixing: two modules with `export * as Contracts from "./origin"` produced origins `a:*` and `b:*` instead of `origin:*`, and the gate reported a collision. `NAMESPACE_REEXPORT` matches *through* `from`, but `specifier_after` required the keyword again, so it never resolved a target and every namespace re-export fell back to its own module. That came in with the origin tracking in `0326ede17` — my own regression, and a false positive that blocks CI for a barrel TypeScript accepts. My existing tests missed it because both paired a namespace re-export against a *declaration*, never two namespace re-exports of one target.

**Commented-out barrel entries.** Right, and the constraint is real: `BARREL_EXPORT` has to run on raw text because stripping blanks a literal along with its quotes, and the module path lives inside them. Since comment-stripping preserves offsets exactly, a match whose span comes back blank was inside a comment. While testing this I found an adjacent pre-existing gap: a *live* entry with a trailing `// note` failed the match entirely, dropping the module from the barrel and surfacing as a confusing orphan-module error. That's accepted now too.

**Visible keyboard hint — the one that mattered most.** This is the finding I'd have wanted caught. The code comment above that paragraph already said "a keyboard *or* screen-reader user has no way to discover the replacement gesture" — and then the fix served only the screen-reader half. A sighted keyboard-only operator (motor impairment, or just someone not using a mouse) never hears `sr-only` text, so the exact person whose Tab key now skips every proof link was the one person who couldn't read why or how to reach them.

The same element becomes visible while focus is inside the grid — announced and shown from one source rather than duplicated — and stays visible while focus moves *between* cells, so it doesn't flash off at the moment the operator starts using the gesture. Two tests cover both: hidden before focus arrives, visible after, and still visible after `ArrowRight` into a cell link.

37 barrel tests, 32 explorer tests including the `jest-axe` pass, strict typecheck clean, and the `tests/scripts` failure set is byte-identical to the pre-existing 9.

One note on process, since this is the seventh round. Every finding for the last four rounds has been in code I wrote earlier in this same PR, and two of them — this one and the credential normalization — were regressions from my immediately preceding commit. That's a real signal about the marginal value of continuing to iterate here versus a human read of the diff. The PRD work has been stable since `d1058650e`; everything after is gate-hardening and this a11y fix. Happy to keep going if that's useful, but I'd suggest a human review pass and a clean `verify-dotnet` run as the better next step.

---

rodoHasArrived:
All four taken — pushed `d37dac02e`. The NaN one is the most consequential finding in this whole review series.

**Non-finite benchmark measurements.** This is a false green in precisely the defect PRD-112 exists to close, and it's nastier than the one originally fixed. An absent result set at least *looks* absent; a NaN row looks measured. Python's JSON parser accepts the non-standard `NaN` token a crashed BenchmarkDotNet run emits, `isinstance(nan, float)` is true, and because NaN is unordered **every** budget comparison returns false — so the stage reported as measured *and* comfortably passing, with NaN written into the evidence file that's meant to be the proof. Measurements must now be finite and non-negative; `bool` is excluded too, since it's an `int` subclass in Python and would otherwise pass as `1.0`. Eleven tests, including one that runs a NaN row through `load_bdn_results` end to end.

**Unicode escapes in skip reasons.** Same family as the two fingerprint collisions fixed last round, and the incomplete decoder was mine: dropping the backslash turned `"A"` into `u0041`, identical to that distinct literal reason though xUnit reports `A`. `\u`, `\U`, and `\x` now decode, with the previously-fixed cases re-asserted so this doesn't reintroduce the chained-replacement bug.

**Single-quoted barrel exports.** Correct that nothing forbids the spelling; the failure mode is also indirect enough to be confusing — the entry isn't discovered, and the *orphan* check then rejects the module as unreachable, which points at the wrong thing. Both quote styles are accepted now, for barrel entries and re-export targets alike.

**Quoted comment before `from`.** Also correct, and it's a false *negative* — the worst kind here. `MODULE_SPECIFIER` searched raw text from the closing brace and found the comment's string, so two modules sharing that comment were assigned one origin and a real collision reported as zero duplicates. The target is now read after the `from` keyword, which `FROM_CLAUSE` locates in comment-stripped text so a `from` inside a comment can't be mistaken for the real one.

40 barrel tests, 52 skip-register tests, 43 budget tests; both gates clean at 1156 names and 19/19 skips; `tests/scripts` failure set byte-identical to the pre-existing 9.

Worth saying plainly, since this is round eight: I'm no longer arguing these findings aren't real — the last two rounds produced four genuine defects, one of which (NaN) would have let a crashed benchmark harness pass as a green performance gate. That's a good catch and I'd rather have it than not. But the same rounds also show the gates are being hardened against increasingly hypothetical inputs while the PRD deliverables have been stable since `d1058650e`, and the one thing actually blocking merge — a clean `verify-dotnet` run past the pre-existing 40 ms flake in `CaptureAsync_PerSourceTimeout_SurfacesTimeoutAfterRetries` — isn't something another review round can resolve.

---

rodoHasArrived:
All four taken — pushed `0886c0b26`.

**File-wide attribute flag → enclosing type.** The strongest of the four. `declares_test_attribute` was computed once per file, so *any* file declaring a `FactAttribute` subclass had every string-valued `Skip =` in it inventoried — an unrelated `new SearchOptions { Skip = "cursor" }` would demand an ownership entry and block a valid change. It's now the *enclosing* type that must derive from `Fact`/`Theory`.

Worth noting what the check confirms rather than just that tests pass: all six genuine cases in this repo assign `Skip` inside the custom attribute's own constructor (`LedgerDatabaseFactAttribute`, `FundAccountDatabaseFactAttribute`, `SecurityMasterDatabaseFactAttribute`, `AssetOperationsDatabaseFactAttribute`, `ReportingDatabaseFactAttribute`, and the two DPAPI attributes). Since the register is keyed on `(path, test, reason)`, any of them dropping out or re-keying would surface immediately as a register mismatch — the gate still reports **19 skips, 19 entries, 0 problems**, so the narrower rule kept every true positive.

**Identity resolved from code only.** Same defect family as the fingerprint collisions, and the same silent-owner-transfer consequence: `resolve_test_identity` scanned raw text, so `// public void Placeholder()` next to a skip was selected as the decorated test and a rename of the real test left the key unchanged. Both identity resolution and the new enclosing-type check now run over a comment- and string-blanked copy with offsets preserved exactly, so scanner positions stay valid. Fixing `read_expression` last round but not this path was an inconsistency on my side — same bug, two places, one fixed.

**`export type *` in the root barrel** — I'd allowed it inside contract modules but not in `types.ts` itself, which is just an inconsistency; a type-only barrel entry would have gone undiscovered and the orphan check would then have blamed the module.

**Destructured exports — rejected, not enumerated.** `export const { Shared } = value` publishes bindings that can nest, rename, default, and rest; enumerating them correctly is parser work, and collecting none of them reports a real collision as zero duplicates. Rejected outright with the explicit alternative, same treatment as a bare star re-export. Zero contract modules use one today.

43 barrel tests, 56 skip-register tests; both gates clean; `tests/scripts` failure set byte-identical to the pre-existing 9.

---

rodoHasArrived:
Both taken — pushed `c92378f83`.

**Prefix-related sibling budgets.** This is the NaN finding's twin: another route to a stage reading as measured when its benchmark never ran, which is the exact defect PRD-112 exists to close. Reproduced before fixing — `_normalise("Bench.ParseTrade")` is `benchparsetrade`, which contains both `parse` and `parsetrade`, so a single result satisfied both budgets and `Parse` evaded the unmeasured gate.

The rule I'd written applied sibling disambiguation only to the *fallback* path, on the assumption that containment was already unambiguous. It isn't, for prefixes. The longest contained stage name is the one a result actually describes, so a shorter sibling contained in the same method name has no claim on it: `Parse` now returns `None` (correctly unmeasured) and `Parse_Trade` claims the result. Sibling context stays optional — called without it, behaviour is unchanged, which the existing `WalChecksum` and `NewlineScan` cases confirm since both go through the fallback path rather than containment.

**Locally re-exported imports.** Correct, and it's a false positive that blocks a valid barrel. `import type { Shared } from "./origin"` followed by a local `export type { Shared }` publishes origin's binding, but my origin tracking attributed it to the module itself, so a sibling republishing that same original binding looked like a second one. Local export clauses now resolve through the module's imports, aliases included (`import { Row as Shared }` binds to origin's `Row`). Both negative directions are pinned: a local clause exporting a *locally declared* name still collides, and two modules importing the same name from *different* sources still collide.

47 barrel tests, 46 budget tests; both gates clean at 1156 names and 19/19 skips; `tests/scripts` failure set byte-identical to the pre-existing 9.

Standing status, unchanged: merge is gated on a clean `verify-dotnet` run past the pre-existing 40 ms flake in `CaptureAsync_PerSourceTimeout_SurfacesTimeoutAfterRetries`, which is identical to `main` and unaffected by any of this.

---

rodoHasArrived:
All three taken — pushed `56794f968`.

**Regex after a control-flow condition.** The consequence is right, though the mechanism runs through a different path than described: the newline bail-out already stopped the scanner blanking to EOF. The actual damage is the brace. With `)` not an opening context, the first slash in `if (value) /\{/.test(value)` read as division, so the regex's `{` reached the brace counter — depth stayed raised for the rest of the file, every later declaration looked nested and dropped out of the comparison, and the gate reported zero duplicates whatever was duplicated. Same false negative, via `{` rather than via blanking.

`)` is now a candidate opening context, admitted only when the literal terminates on the same line. That's an invariant rather than a heuristic — a JavaScript regex literal cannot contain an unescaped newline — and checking *before* blanking means a misread slash costs nothing, where the old shape blanked as it scanned and only then discovered the newline. Verified directly: brace depth stays 0, the later export survives, and `const q = (a + b) / c;` is untouched. `(a) / b / c` is also covered, since two slashes on one line are the case that could over-trigger.

While splicing this I left a `continue` that would have infinite-looped on a non-regex slash after `)` — caught by running the suite rather than by reading it.

**Verbatim and raw skip literals.** A real fingerprint collision of the same family as the earlier ones: `@"A   B"` and `@"A B"` bypassed the pure-literal path, fell through to `normalise_expression`, and had their interior whitespace collapsed into one key. Both forms now decode to exact content — doubled quotes handled, and no backslash escapes applied, since `@"C:\network"` is a literal backslash-n.

**Undecodable sources — fail closed.** The clearest of the three. A `UnicodeDecodeError` dropped the whole file, so every skip in it was both unregistered *and* unreported while the gate announced that all skips are owned. That's the one failure mode a coverage claim can't survive. UTF-16 and UTF-8 BOMs are now honoured; anything still undecodable raises `GateFailure`. Tested both ways — a UTF-16 source is read, and invalid UTF-8 with no BOM fails the gate instead of vanishing.

50 barrel tests, 63 skip-register tests, 46 budget tests; both gates clean at 1156 names and 19/19 skips; `tests/scripts` failure set byte-identical to the pre-existing 9.

---

rodoHasArrived:
Both taken — pushed `d305003c6`. The regex one is a regression I introduced last round.

**Quote inside a regex probe.** Admitting `)` as an opening context — my fix for the previous round's brace-leak — created this: `(total) / "x/y".length` let the probe terminate on the slash *inside* the string, blank through it, and leave the string's closing quote to open another one, swallowing any declaration later on that line. Reproduced before fixing, and the same-line placement matters: with the export on the *next* line it survives, which is why my own test for the previous fix didn't catch this.

A quote before the closing slash now rejects the candidate, falling back to reading the slash as division — exactly what this position meant before regexes were admitted here. Regexes matching a quote character exist but are vanishingly rare in contract modules, and the fallback is the safe direction. All three cases verified together: the string case, `if (value) /\{/.test(value)` on one line, and plain `(a + b) / c`.

**A backstop, because this is now twice in two rounds.** Both misclassifications were silent in the same way — brace depth decides which declarations are at module scope, so a misread leaves later declarations looking nested and the gate reports zero duplicates from a truncated view. A hand-rolled lexer will keep producing these; what it shouldn't do is report confident results from a parse it botched. A module whose stripped text has unbalanced braces is now reported rather than trusted. I verified all 14 contract modules parse balanced today, so this bounds the failure mode without changing any current result.

That doesn't make the lexer correct, and I want to be straight about it: this class of finding is not exhausted. The backstop converts the *brace-corruption* variety into a loud failure, but a misread that happens to stay balanced — as the string case above did — still slips through. A real TypeScript parse is the only thing that closes it properly, which is out of reach for a Python CI script.

**`\e`.** C# 13's ESC escape was missing from the table, so `"\e"` decoded to the string `"e"` — the same fingerprint as the distinct reason `"e"`. Added, with the collision pinned.

53 barrel tests, 64 skip-register tests; both gates clean at 1156 names and 19/19 skips; `tests/scripts` failure set byte-identical to the pre-existing 9.

---

PR #2539, opened 2026-07-28
Deliver portfolio-aware risk engine on the existing order-gate rail

rodoHasArrived:
## Summary

Meridian.Risk promised a pluggable, severity-graded, portfolio-aware pre-trade architecture but delivered three rules and ~250 LOC. This PR closes that promise/delivery gap on the existing order-gate rail — deliberately not the deferred enterprise-risk lane (no VaR, no Greeks) — and composes with W9-SAFETY-007 (kill-switch/fat-finger/collar) rather than duplicating it.

**PortfolioExposure fed from the existing `AggregatePortfolioService`.** New `IPortfolioExposureProvider` + `PortfolioExposureSnapshot` in Meridian.Risk, adapted from `IAggregatePortfolioService` in Ui.Shared so enforcement evaluates the same aggregated cross-run exposure the Portfolio workspace reports. `PortfolioRegistry` and `IAggregatePortfolioService` are now actually registered in DI (they existed but were never wired — the `/api/workstation/portfolio/aggregate|exposure|symbol-exposure` endpoints returned 503); the host paper portfolio pre-registers as `workstation-paper`. The F# `RiskContext.PortfolioExposure` field — designed-in but never fed — is now populated.

**The exposure/concentration/notional rules the context was designed for**, with policy in the F# risk domain like the existing rules:
- `GrossExposureRule` (severity **Critical**) — projected portfolio-wide gross exposure vs. ceiling.
- `SymbolConcentrationRule` (severity **Error**) — projected single-symbol % of portfolio value vs. cap, with an approved-with-warning observe band at ≥80% of the cap.
- `OrderNotionalRule` (severity **Error** + escalation band) — hard ceiling rejects; at/above the escalation band the previously-unused F# `Escalate` decision parks the order.

**Severity now has real outcomes** in `CompositeRiskValidator`:
- `Info`/`Warning` → breach becomes a warning flag; evaluation continues.
- `Error` → rejects (unchanged).
- `Escalate` → parks the order in the new `RiskEscalationQueueService` for governed approval. Operators approve/deny with audited actor+reason; approval arms a one-shot, fingerprint-matched release token (order metadata `riskEscalationId`) that passes exactly that order through the escalation while every other rule still enforces. No queue → fails closed as rejection.
- `Critical` → rejects **and trips the existing execution circuit breaker** (`ExecutionOperatorControlService`), halting routing until an operator closes it. The drawdown guardrail was already declared Critical — it now genuinely trips the breaker it is named after.

[...]

PR #2557, opened 2026-08-01
Apply nine outstanding review findings to the W10 rows

rodoHasArrived:
## Summary

Rounds 12 and 13 of review on #2548 landed while that PR was mid-merge. All nine findings were verified against source and none were rebutted, but applying them would have superseded the commit the merge was gated on. They land here instead, against the merged rows.

**Round 14 arrived on this PR** and is applied in the second commit — see below.

### Five criteria corrections

**`W10-RECON-002` — materiality is a per-break property.** The classifier sets `IsMaterial` from each break's absolute variance. A group whose *signed aggregate* sits below a group threshold can still contain a material member, and gating on the aggregate lets that member bypass the independent approval it individually requires. Approval now triggers whenever any member is material or high-risk, and a group cap is measured on **gross** exposure rather than a net that can offset one away.

**`W10-SEAM-001` — present and fresh is not the same as about the same thing.** An earlier round made an absent contributor a blocking state. This is the other half: contributors can all be registered, healthy, and current while describing *different subjects*. Every contribution must now answer for the same fund profile, ledger book, fund account, entity, and period, with a mismatch blocking the projection rather than merging into it.

**`W10-TAX-001` — an open replacement window makes zero exposure provisional, not settled.** The wash-sale window is symmetric, so an acquisition *after* the disposal can still change the disallowed loss. The engine computes zero from acquisitions known today and the generic incomplete-input rule does not catch it — the inputs are complete, the window simply hasn't closed. Now requires a provisional label, the remaining window, and re-evaluation or governed finalization on close.

**`W10-RECON-001` — absence from a run is evidence of clearing only when that run concluded.** Corrected in round 14; see below.

**`W10-JRNL-001` — idempotent posting does not imply idempotent drafting.** The due-occurrence calculation keeps returning an occurrence until it is confirmed posted, so a runner polling while the first draft awaits approval enqueues another one every pass. The claim is now taken at initial enqueue, keyed by schedule and occurrence date, with retries returning the existing draft.

### Round 14 — three findings applied, one partly rebutted

**Applied — the diff-successor status set was inverted.** Copilot and Codex found this independently and source confirms both. `StatementRunStatus` (`StatementReconciliationDtos.cs:9-20`) carries `ReviewRequired = 6` *and* `Completed = 7` as distinct successful terminal states, and `ReconciliationApiService.cs:92` assigns `breakDtos.Length == 0 ? Completed : ReviewRequired` (same shape at `:187-188`, `:232`, `:634`). `Completed` is therefore **exactly the set of runs with zero breaks** — so gating the diff on it would suppress run-over-run comparison precisely when breaks exist. Codex's case is the sharp one: where break A remains and break B disappeared, B never clears. The criterion now names a *successfully reconciled terminal run* (`ReviewRequired` or `Completed`) and excludes only `ValidationFailed`, `Failed`, and `Canceled`.

**Applied — module routing was one-directional, and more broadly than reported.** `source_modules` renders on the roadmap side, but `render-source-docs.py:60-70` builds source-roadmap traceability **only** from each module's reverse `roadmap_items` list in `docs/source/data/source-modules.yml`. Those lists are actively maintained — W1 through W9 are present — but held **no W10 entries at all**. Adding just the four corrected links would have left the registry selectively populated, which reads worse than uniformly empty. All eleven rows are now linked under every module they claim: **63 entries across 11 modules**.

**Partly rebutted — the inaccuracy finding named a different service.** The note said "the read service" without naming it, which is what let it be resolved to `FundOperationsWorkspaceReadService`. That service does scope by fund, exactly as the review says. But it is not the service the finding traced. `FundLedgerViewModel.cs:1037-1038` calls `IFinancialOperationsCommandCenterReadService.GetCommandCenterAsync(fundProfileId: …)` with **only** a profile id; `FinancialOperationsCommandCenterReadService.cs:42-43` then calls `ListAsync(fundAccountId, periodId, status: null, ct, ledgerBookId: ledgerBookId)` with all three null — listing across every fund — picks one via `ResolveActiveWorkflow`, adopts its fund at `:57`, and queries the cockpit with the requested profile at `:66`. **The original finding stands**; the prose defect was mine, and both services are now named explicitly.

The review's own observation is kept as a *second* constraint: `FundOperationsWorkspaceReadService` selection ignores the summary's `LedgerBookId` and `PeriodId`, giving a right-fund, wrong-book-or-period mismatch. The amended criterion demands binding on both, so both services need it.

### Where the mechanism went

Each correction's supporting detail is added to **Known Constraints from Source** in `docs/product/w10-depth-slate-2026-07.md`, attributed to the type that owns the behaviour — not into the criteria. That section is blueprint input; the criteria state outcomes.

## Reason

These are the last unapplied findings from thirteen review rounds on #2548, plus round 14 on this PR. Leaving them would mean merged rows carry acceptance criteria that a reviewer has already shown to be satisfiable incorrectly.

## Testing performed

- [x] `main` merged in (PR #2555 landed 20 commits mid-review); three generated-dashboard conflicts resolved to main's side and regenerated
- [ ] Relevant unit tests were added or updated — n/a, registry data change

```
python3 build/scripts/docs/validate-roadmap-registry.py --summary   # 0 errors, 0 warnings
python3 build/scripts/docs/render-roadmap-docs.py --summary         # 2 changed
python3 build/scripts/docs/render-roadmap-diagrams.py --summary     # 0 changed
python3 build/scripts/docs/render-source-docs.py --summary          # 13 changed
python3 build/scripts/docs/run-docs-automation.py --profile core ...  # all success
python3 scripts/generate_program_state_summary.py
python3 scripts/check_program_state_consistency.py                  # passed
```

39 rows; all eleven `W10-*` inside the 3–6 outcome-criteria band. Source-roadmap traceability now carries 63 W10 links where it previously carried none.

**Phase marker moves to `phase:PR9`.** Regenerating source docs rewrites the per-module `src/**/README.md` traceability tables, which `PR1` does not cover. `tools/roadmap/enforce_phase_scope.py --phase PR1` exits 1 on those files; `--phase PR9` exits 0. Read by exit code, not through a pipe — piping substitutes the pipe's status and hides the failure.

## Safety review

## Governance changes


rodoHasArrived:
Round 14: three of four applied in `96f9325`, one partly rebutted. Also merged `main` (PR #2555 landed 20 commits mid-review) — the three conflicts were all generated dashboards, resolved to main's side and regenerated.

**Applied — the `Completed` collision.** Confirmed against source, and Copilot and Codex found it independently. `StatementRunStatus` has `ReviewRequired = 6` and `Completed = 7` as two distinct *successful* terminal states, and `ReconciliationApiService.cs:92` assigns `breakDtos.Length == 0 ? Completed : ReviewRequired`. So `Completed` is exactly the zero-break set, and my wording pointed an implementer at the one status that excludes every run the diff exists to compare. Codex's framing is the sharpest: break A remains, break B disappeared, B never clears. Fixed by naming a *successfully reconciled terminal run* and excluding only `ValidationFailed`, `Failed`, `Canceled`.

**Applied — reverse module routing, wider than reported.** Right that `render-source-docs.py:60-70` builds traceability only from `source-modules.yml`'s reverse `roadmap_items`. Worth adding: those lists carry W1–W9 but had **zero** W10 entries, so the gap was all eleven rows, not the four corrected links. Adding only four would have left the registry selectively populated — arguably worse than uniformly empty. All eleven are now linked under every module they claim: 63 entries across 11 modules, and the traceability table went from 0 W10 rows to 63.

**Partly rebutted — the inaccuracy finding targets a different service.** `FundOperationsWorkspaceReadService` does scope by fund, exactly as described (`:2600-2614`). But that is not the service the finding traced. The path is:

- `FundLedgerViewModel.cs:1037-1038` → `IFinancialOperationsCommandCenterReadService.GetCommandCenterAsync(fundProfileId: activeFund.FundProfileId, ct: ct)` — only the profile id; fund account, ledger book, period, entity all null.
- `FinancialOperationsCommandCenterReadService.cs:42-43` → `ListAsync(fundAccountId, periodId, status: null, ct, ledgerBookId: ledgerBookId)` with all three null, so the listing spans every fund.
- `ResolveActiveWorkflow` picks one; `:57` adopts its fund via `effectiveFundAccountId = fundAccountId ?? activeWorkflow?.FundAccountId`; `:66` queries the cockpit with the *requested* profile.

So the cross-fund combination is real and the original finding stands. The defect was in my prose — it said "the read service" without naming it, which is precisely what allowed it to resolve to the wrong type. Both services are now named with call sites.

The observation in the finding is kept as a genuine *second* constraint, because it is a real defect I had not recorded: `FundOperationsWorkspaceReadService` selection ignores the summary's `LedgerBookId` and `PeriodId`, giving a right-fund, wrong-book-or-period mismatch. The amended `W10-SEAM-001` criterion demands binding on both, so both services need work.

**One process note.** The phase marker moves `PR1` → `PR9`. Regenerating source docs rewrites the per-module `src/**/README.md` traceability tables, which `PR1` does not cover — `enforce_phase_scope.py --phase PR1` exits 1 on them, `--phase PR9` exits 0. Caught by reading the exit code directly rather than piping it.

---

rodoHasArrived:
Round 15: all three applied in `f23bffa8`, each confirmed against source first. Main merged too (60 commits — #2539 and #2558); the generated-dashboard conflicts resolved to main's side and regenerated.

**Both profile pairs — confirmed and sharper than I'd written.** `StatementRunDto` (`StatementReconciliationDtos.cs:277-282`) carries `MappingProfileId`/`MappingProfileVersion` *and* `ToleranceProfileId`/`ToleranceProfileVersion` as independent nullable members, and `ReconciliationApiService.cs:693-696` hardcodes **both versions to `null`** while passing the ids through. So a nullable equality check compares `null == null` and reads two materially different runs as matching — the criterion would have been satisfiable by an implementation that checks nothing. Now requires both profiles at a known matching version, with an absent version treated as unknown rather than as agreement. Recorded in Known Constraints that the adapter has to start writing them for the criterion to be satisfiable at all.

**Provisional narrowing — confirmed, including the caveat.** `ComputeWashSale` (`LedgerTaxLotReliefProjector.cs:225-232`) returns `null` when `realizedGainOrLoss >= 0m`, so a gain-only disposal cannot attract a wash sale however the window closes, and my blanket rule would have forced re-evaluation that can never change the answer. The exception you name is real and is in the method's own doc comment: a *net* gain containing loss shares is not yet decomposed to defer only those shares, so it can't be excluded today. The criterion now covers loss-generating disposals and mixed gain/loss reliefs, and lets a wholly-gain disposal read as settled.

**Canonical close scope — confirmed, and it supersedes what I wrote last round.** Every dimension of `GetCommandCenterAsync` is optional (`FinancialOperationsCommandCenterReadService.cs:32-43`), and `:56-58` fills gaps from the selected workflow — `effectiveFundAccountId = fundAccountId ?? activeWorkflow?.FundAccountId`, same for period. Requiring contributors to agree *with each other* is the weaker property: a null dimension cannot register as a mismatch because there is nothing to compare it against, so a recently-updated workflow for the wrong book or period still becomes the "consistent" subject. The criterion now puts the obligation on the request or an explicit operator selection to establish the complete scope, with missing or ambiguous dimensions blocking.

39 rows; all eleven W10 rows still inside the 3–6 outcome-criteria band. Registry validates 0/0, and the full `regenerate-docs` sequence was rerun against the merged tree.

---

rodoHasArrived:
**`schema-control` is failing here, and I do not think it is this PR's to fix.** Flagging rather than pushing a speculative manifest.

The failure is `Committed schema-control manifests or documentation are stale.` from `schema-control.py verify --base-ref origin/main`.

Why this PR trips it: the workflow's path filter includes `src/Meridian.Contracts/**`, and `render-source-docs.py` regenerates `src/Meridian.Contracts/README.md` to carry the new W10 traceability rows. That README is my **only** change under any watched path — the full diff of this branch against `main` for `src/Meridian.Contracts/`, `database/`, `docs/generated/database/`, and `docs/reference/database-schema.md` is exactly that one generated markdown file. No `.cs` contract, no migration, no database artifact.

Why the staleness is pre-existing: `schema-control.yml` triggers on `pull_request` only — `opened`, `synchronize`, `reopened` — with **no `push:` trigger**. It never runs on `main`. So when #2539 merged 60 commits of risk/portfolio contract work, nothing regenerated or verified the committed manifests, and nothing would have. This PR is simply the first since then to touch a path under `src/Meridian.Contracts/**`, so it is the first to surface the condition.

I cannot regenerate the manifests here: `snapshot` and `verify` both build a candidate database and need PostgreSQL, which this environment does not have (`inventory` is the only DB-free mode and covers migrations, not the contract manifest). Producing manifest content by hand would be guessing at generated output, which is worse than leaving it visible.

Two ways forward, both outside this PR:

1. Run `schema-control.py snapshot` then `promote` against current `main` and commit the refreshed manifests — the check would then pass here on the next synchronize.
2. Consider adding a `push:` trigger on `main` to `schema-control.yml`, so this class of staleness is caught at merge rather than surfacing on whichever unrelated PR next touches a watched path. Right now a docs-only change is carrying the signal for a contracts change that landed separately.

Everything else on this PR is green or in flight, and the sibling PR #2561 — which does not touch a watched path — is fully green including `verify-dotnet`, so the rest of the pipeline is healthy.

---

rodoHasArrived:
**Round 16 applied** — `a6652a1ed`.

Confirmed against source. `ComputeWashSale` short-circuits on `!input.WashSalePolicy.AppliesOn(input.SaleDate)` at `LedgerTaxLotReliefProjector.cs:232`, *before* it reaches the `realizedGainOrLoss >= 0m` test that round 15 addressed. `AppliesOn` is `Enabled && (EffectiveDate is not { } effective || saleDate >= effective)` (`WashSale.cs:66-67`), and `WashSalePolicy.Disabled` (`:47`) is the named default. `PostgresLedgerJournalStore.WashSale.cs:26` gates on the same predicate, so the exclusion holds on the store side too.

So an account with no wash-sale policy configured, or one whose policy takes effect after the sale date, can never accrue exposure however the replacement window moves — labelling it provisional promises a re-evaluation that has nothing to re-evaluate. Governance is now the first test, ahead of the loss test: provisional belongs to a disposal a policy actually governs on its own sale date that either realizes a loss or holds mixed gain and loss lots. Everything else reads as settled.

Mechanism recorded under **Known Constraints from Source**, attributed to `WashSalePolicy` and `LedgerTaxLotReliefProjector`. Registry still 39 rows, `W10-TAX-001` still 5 criteria, validator 0 errors / 0 warnings. Full regeneration sequence rerun after merging `main` (which had moved again).

`schema-control` remains red for the inherited reason described [above](https://github.com/rodoHasArrived/Meridian-main/pull/2557) — the branch's only change under a watched path is a generated `src/Meridian.Contracts/README.md`, and refreshing the manifests needs PostgreSQL.

---

rodoHasArrived:
**Round 17 applied** — `9273eb32b`. Both findings confirmed against source; neither rebutted.

### `W10-RECON-001` — a failed run can present as a successful one

This is genuinely new evidence beyond round 14, and it inverts what that round concluded was sufficient. Round 14 established that `Completed` is exactly the set of runs with zero breaks, and treated the `ReviewRequired`/`Completed` pair as the conclusive-run test. The save-before-reconcile path shows that pair is not enough:

- `StatementRunWorkflowService.CreateAsync` persists the import through `brokerStatementService.ImportAsync` **before** it calls `_populationProvider.GetPopulationsAsync`, `StatementRunMatcher.Match`, or `breakStore.WriteAsync`. A throw anywhere in that later stretch leaves a durable import with **no breaks attached**.
- `ReconciliationApiService.cs:634` — `var status = openExceptionCount > 0 ? StatementRunStatus.ReviewRequired : StatementRunStatus.Completed;` — labels any retained import by counting open exceptions.

So a run that never reconciled reads as `Completed`, and lineage gated on the terminal-status pair would clear **every** prior break on the strength of it. The criterion now requires the successor to have **durably recorded that reconciliation itself completed**, and explicitly refuses a status inferred from an absence of breaks. The ordering guard already in `CreateAsync` — resolving the tolerance profile before `ImportAsync` precisely so a bad profile cannot strand an import — is the same concern one step earlier, and shows the codebase already treats save-before-reconcile as a hazard.

One note for whoever implements it: this makes recording the outcome a prerequisite, alongside recording the two profile versions that round 14 established are currently hardcoded to `null`.

### `W10-RECON-002` — gross exposure over incomparable amounts

Confirmed. `ReconciliationBreakQueueItem.Variance` (`Workstation/ReconciliationDtos.cs:407-414`) is a bare `decimal` — no currency, no unit. The typed measures beside it, `ReconciliationBreakMeasureDto` (`:376-383`), carry `Kind` of `Value` / `Quantity` / `CostBasis` plus a free-text `Unit`, and the queue projection can place a quantity delta into that same top-level variance. Summing absolute variances across a cause group can therefore add shares to cash or USD to EUR, and trip or skip the threshold on a number that means nothing.

Worth saying that this does not weaken round 13's gross-over-net finding — it constrains it. Gross was the right measure; it just needs a commensurable domain. Exposure is now aggregated only within a comparable measure and currency, or across a governed translation, and a member whose unit is unknown or whose rate is unavailable **holds the group at the stricter approval** rather than being summed into a meaningless total. That keeps the fail-closed direction consistent with rounds 15 and 16.

Both mechanisms recorded under **Known Constraints from Source** with their owning types. Registry 39 rows, both rows inside 3–6 criteria, validator 0 errors / 0 warnings, full regeneration sequence rerun.

`schema-control` stays red for the inherited reason reported earlier — the watched-path diff against `main` is still exactly one generated `src/Meridian.Contracts/README.md`, and refreshing the manifests needs PostgreSQL.

---

rodoHasArrived:
`schema-control` is red on `cb57a33849`, the current head. Still the inherited failure, but I dug further this time and have two things worth more than the previous "needs PostgreSQL" summary.

### 1. This PR trips the check on a file that cannot cause schema drift

The watched-path diff against `main` is still **exactly one file**, and now that I have looked at its contents it is clearly unrelated to the database:

```
 src/Meridian.Contracts/README.md | 10 ++++++++++
```
```diff
 <!-- source-roadmap-traceability -->
+| `W10-MARK-001` | Fail-closed stale-mark policy and mark-age surfacing |
+| `W10-RECON-001` | Durable break lineage identity and run-over-run break diff |
   … eight more W10 row titles …
 <!-- source-roadmap-traceability:end -->
```

That is `render-source-docs.py` appending roadmap row titles to a generated traceability table. No DTO, no migration, no schema. The workflow's path filter includes `src/Meridian.Contracts/**` (`schema-control.yml:15`), which is right for contract *types* but also matches a **generated markdown README** that is structurally incapable of affecting a PostgreSQL schema.

So the durable fix is a path-filter refinement — exclude `**/*.md`, or at least the generated READMEs — which would stop pre-existing manifest staleness from blocking unrelated docs PRs. I have deliberately **not** made that change here: `.github/workflows/**` is on the AGENTS.md protected list, this PR is not designated for human governance review, and folding a workflow change into a roadmap-registry PR would be scope creep. Flagging it as the real fix rather than doing it unasked.

### 2. Refreshing the manifests does not need a local PostgreSQL after all

I previously said this needs a database this environment lacks. That is true of *my* environment but not of the repository — **the workflow provisions PostgreSQL itself**:

- `services.postgres: postgres:16.14-alpine` with `DATABASE_URL` already wired (`schema-control.yml:39-40`, `:49-63`)
- `workflow_dispatch` accepts `mode: snapshot`, gated at `:88` to run `schema-control.py snapshot`
- every run uploads `build/schema-control/` as an artifact (`:110-117`)

So the path to green does not require anyone to stand up a database:

1. Actions → **PostgreSQL Schema Control** → *Run workflow* on `main` with `mode=snapshot`
2. download the `schema-control-<run>-<attempt>` artifact
3. `schema-control.py promote` the reviewed candidate into the committed outputs and commit that to `main`

Step 3 is the only part needing a human — the job runs with `permissions: contents: read`, so it cannot commit its own output, by design.

I have not triggered that dispatch myself: it is an outward-facing action on your CI, and the result still needs someone to review and commit the refreshed manifests. Say the word and I will run it and prepare the commit.

The underlying cause is unchanged from what I reported earlier — `schema-control.yml` has **no `push:` trigger**, so it never runs on `main` and nothing regenerated the manifests when #2539 merged 60 commits of contract work. A `push:` trigger on `main` would catch this class of staleness at merge instead of surfacing it on whichever unrelated PR next touches a watched path.

Everything else on this PR is green, and rounds 16 and 17 are applied.

---

rodoHasArrived:
**Round 18 applied** — `00d3fbe93`. Both findings confirmed against source; neither rebutted.

### `Completed` is not the zero-break set

Confirmed, and my round-14 wording was wrong in a way worth naming precisely: I wrote that four sites share the shape `breakDtos.Length == 0`, and they do not.

| Site | Test | Meaning |
| --- | --- | --- |
| `ReconciliationApiService.cs:92` | `breakDtos.Length == 0` | zero break **records** |
| `:232` | `breaks.Length == 0` | zero break **records** |
| `:186-188` | `!breaks.Any(item => item.Status == "Open")` | zero **open** breaks |
| `:634` | `ResolveOpenExceptionCount(...) == 0` | zero **open** breaks |

So the status a caller receives from `GetStatementRunAsync` or from a summary projection also covers runs that found breaks and resolved every one of them. Round 14 generalised from the two record-count sites and asserted the property for all four.

The consequence you name is the sharp one, and it is worse than a documentation inaccuracy: a run whose breaks all resolved is the *most* informative successor there is — it is where clearing actually happened — and status alone cannot distinguish it from a run that produced no breaks at all. An implementation that read `Completed` as "no breaks existed" would discard exactly the resolved historical occurrences the run-over-run diff exists to preserve. The evidence bullet is now split in two: the terminal-status point stands on its own, and a new bullet states the four-site divergence and requires lineage to read break records rather than run status.

`W10-RECON-001` itself is unchanged, deliberately. Its second criterion already carries the property in outcome form — a recurrence must remain "a distinct occurrence" and must not "overwrite the interval during which it was clear", which is unsatisfiable without retaining resolved occurrences. Adding a sixth criterion for something already required would push the row past the W9 house-style band without adding an obligation.

### Source README hash entries

Confirmed, and slightly wider than reported: the branch regenerates **eleven** source READMEs, not six — `SRC-APP`, `SRC-BACKTESTING`, `SRC-CONTRACTS`, `SRC-DOMAIN`, `SRC-FSHARP`, `SRC-DESIGN-FINANCIAL-OPERATIONS`, `SRC-LEDGER`, `SRC-STRATEGIES`, `SRC-UI-SHARED`, `SRC-UI-DASHBOARD`, `SRC-WPF`.

Refreshed with repeated `--write-module`, not `--write`. The distinction matters here: the validator reports 53 findings on this commit, and only 22 of them belong to these eleven modules. The other 31 are inherited drift on modules this branch never touches, and `--write` would have rewritten those entries too — recording an alignment review that did not happen. After the scoped refresh: **53 → 31**, with no README drift left on any of the eleven.

Verified: docs automation core profile clean, `test_roadmap_source_docs.py` 13/13, AI inventory 322 items 0 findings, orphan count still 255, registry validator exit 0, every changed path exits 0 against `enforce_phase_scope.py --phase PR9`.

### `schema-control` — new evidence, and a correction to my own recommendation

I pulled the job log and the artifact this time.

**The tooling is healthy; only the comparison fails.** PostgreSQL 16.14 starts cleanly, the 67 schema-control tooling tests pass (`Ran 67 tests in 0.149s / OK`), migration errors 0, policy errors 0. The single failing step is `verify`, on `Committed schema-control manifests or documentation are stale.`

**The drift is bounded and clearly attributable.** Artifact `schema-control-160-1` (81 files) reports **15 changed, 0 added, 0 removed** — `manifest/{catalog,contracts,dependencies,migrations,policies}.json`, `manifest/schemas/{ledger,reporting}.json`, and eight generated docs under `ledger`/`reporting`. Diffing the candidate against the committed tree shows what is missing: `ledger` gains `ledger_wash_sale_deferrals` and four `wash_sale_*` columns on the policy table; `reporting` goes 15 → 20 relations with `reporting_reporting_run_create_claims`, `reporting_reporting_run_snapshots`, and a `consumed_artifact_ids` column. All of that is already-merged migration work that no run ever regenerated, because the workflow has no `push:` trigger.

**Correcting what I proposed last time:** the `workflow_dispatch mode=snapshot` step is unnecessary. `verify` and `snapshot` call the same `create_snapshot(...)` with identical arguments (`cli.py:749-760`) — `snapshot` just returns early, while `verify` continues to compare. So the candidate this PR's *failing* run already produced and uploaded is byte-equivalent to a snapshot run. And `promote` short-circuits at `cli.py:744-747` before `create_snapshot`, so **it needs no database at all**. The recovery is: download artifact 8836916761 → `schema-control.py promote --candidate-root <extracted>/candidate` → commit.

The candidate is `main`-equivalent, which I checked rather than assumed: the diff of this branch against `main` across *every* watched path is one file, `src/Meridian.Contracts/README.md` — ten generated roadmap-traceability rows — and `database/` is empty.

I tried to run that promote on this branch and **the sandbox permission classifier blocked it** — the copy into the working tree and the `promote` invocation were both denied. So I have the correct output in hand but cannot commit it. If you want it on this PR, allowing the promote command would let me finish it in one step; the result is self-checking, since a wrong promote simply leaves the next `verify` red. Otherwise it belongs on `main` directly.

The two durable fixes are unchanged: a `push:` trigger on `main` so this is caught at merge, and narrowing the `src/Meridian.Contracts/**` path filter to exclude `**/*.md` so a generated README stops carrying the signal for contract work that landed elsewhere. I have not touched `schema-control.yml` — it is on the AGENTS.md protected list and this PR is not designated for governance review.

---

rodoHasArrived:
**Round 19 applied** — `c4b5eb907`. All three confirmed against source; none rebutted.

### `W10-RECON-002` — the fail-closed branch I added could never fire

This is the sharpest of the three, because it shows round 17's fix was inert for the case it most needed to cover. That round required exposure to be aggregated only across "the same measure and the same currency", with an unknown unit holding the group at the stricter approval. For quantity, the unit is a **constant**:

- `ReconciliationBreakQueueProjection.BuildStatementBreakMeasures` (`:223-263`) stamps the literal string `"units"` on every quantity measure, whatever the instrument. Not derived, not nullable — a string literal in the constructor call.
- `ReconciliationBreakQueueItem` (`Workstation/ReconciliationDtos.cs:407-`) carries no instrument dimension at all. `ExternalAccountId`, `CustodianId`, `Counterparty`, and `FundAccountId` are the closest it comes; there is no security id and no symbol.

So 100 shares of one security and 100 shares of another are the same `Kind`, in the same `Unit`, and the unit is never *unknown* — which means the escape hatch round 17 added never triggers, and the criterion reads as satisfied while summing incommensurable quantities. The rule now splits by measure: value exposure within one currency or a governed translation, quantity exposure only across members naming the same instrument under a governed quantity unit. Recorded that this makes an instrument dimension on the queue contract a prerequisite, not an optional refinement — until it exists, a cause group mixing instruments holds at the stricter approval.

Worth being explicit that this is the third round in a row where a criterion promised a re-check the engine cannot perform. The pattern is the same each time: I write a general rule, and the specific shape of the data makes the general rule vacuous.

### `W10-JRNL-001` — schedule-version drift

Confirmed, and your framing of why the obvious fix is wrong is what makes it useful. `FundAdministrationControlService.ScheduleRecurringJournal` (`:105-130`) overwrites in place — `_recurringSchedules[schedule.ScheduleId] = schedule;` — and `RecurringJournalSchedule` (`src/Meridian.Ledger/RecurringJournalSchedule.cs:22-36`) has no version member. So re-scheduling one id with a changed template, ledger book, or parameters leaves the schedule/date claim resolving to a draft built from the superseded definition, and returning it applies stale content silently. Putting the version *in the claim key* would trade that for two approval drafts covering one logical occurrence — the exact duplicate the claim exists to prevent. The criterion now has the claim **retain** the schedule and template versions and fail closed on drift until the existing draft is explicitly superseded or corrected.

One supporting detail: the codebase already names this hazard one level up. The template-registration path at `FundAdministrationControlService.cs:80-82` comments that "recurring schedules resolve the current template at materialization, so the chain must preserve which version was approved" — and hashes the template definition into its registration event for exactly that reason. The schedule overwrite two methods below has no equivalent guard.

### `W10-TAX-001` — the disallowance saturates

Confirmed. `ComputeWashSale` (`LedgerTaxLotReliefProjector.cs:246-259`) caps the match at the quantity sold:

```csharp
var matchedQuantity = Math.Min(input.QuantitySold, matching.Sum(r => r.Quantity));
var disallowedLoss  = RoundCurrency(totalLoss * (matchedQuantity / input.QuantitySold));
disallowedLoss      = Math.Min(disallowedLoss, totalLoss);
```

Once known replacements cover the whole quantity sold, the ratio is already 1, `disallowedLoss == totalLoss`, and `allowedLoss == 0`. A later acquisition inside the remaining window cannot move either number. Labelling that disposal provisional promises a re-evaluation with nothing left to re-evaluate — the same defect as the governing-policy narrowing in round 16 and the gain-only narrowing in round 15, in a third form. Provisional is now scoped to disposals whose known replacement acquisitions do not yet cover the entire quantity sold; a fully matched one reads as settled alongside the ungoverned and gain-only cases.

### Verification

Registry 39 rows, all eleven W10 rows inside the 3–6 band, validator exit 0, AI inventory 322 items / 0 findings, orphan count still 255. One extra step this round: the amendments tripped `docs/roadmap/generated/MANIFEST.json`'s inputs hash, so `render-roadmap-docs.py` was rerun — doc-hash findings are back to the 31 inherited, not 32. Every changed path exits 0 against `enforce_phase_scope.py --phase PR9`.

`schema-control` is unchanged from the previous comment: the candidate from the failing run is in hand and `promote` needs no database, but the sandbox blocked the command. That one still needs you.

---

rodoHasArrived:

rodoHasArrived:

PR #2560, opened 2026-08-01
Stop presenting the local clock as a market observation time

rodoHasArrived:
## Summary

Two WPF surfaces presented a local clock reading as a market observation time. In a product where mark age is a governed control, that is a false reading rather than a cosmetic one.

### 1 — `PositionBlotterViewModel`: no observation time exists, so say so

`MapToEntry` stamped `MarketTime` from `DateTime.Now`, so the **"Min mkt time"** and **"Max mkt time"** tiles reported the moment the grid was constructed and labelled it as when the market was observed.

**There is no real value to substitute, and nothing is being dropped in projection.** Verified against source:

| Type | Observation timestamp? |
| --- | --- |
| `ExecutionPositionDetailResponse` (`ExecutionApiModels.cs:17-37`) | none — `Expiration` is a contract expiry |
| `ExecutionBlotterSnapshotResponse.AsOf` (`:6-12`) | `DateTimeOffset.UtcNow` at response construction (`ExecutionEndpoints.cs:1110`, `:1134`) — same class of value |
| `BrokerPosition` (`IBrokerageGateway.cs:175-224`) | no timestamp member |
| `IPosition` (`IPosition.cs:16-41`) | no timestamp member |

The observation time does not exist anywhere in the chain, so the honest fix is to leave `MarketTime` unset. `BlotterEntry.MarketTime` is already `TimeOnly?` (`BlotterModels.cs:72`) and `UpdateStatusBar` already degrades correctly — `:595` filters on `HasValue`, `:606-607` renders `"—"`. No model change is needed to express "unknown", and `"—"` is true where the current value is not.

Plumbing a genuine observation time through the contract belongs to **`W10-MARK-001`**, where it is already recorded.

### 2 — `LiveDataViewerViewModel`: the real timestamp was parsed, then thrown away

`LastTradeTimeText` came from `DateTime.Now` even though the trade's **own timestamp is already parsed and retained** as `RawTimestamp` and then discarded. It is now carried alongside `_lastPrice` and rendered instead — which also makes the tile **agree with the events grid**, which has always rendered the same raw value.

Getting that right took three attempts, and the two follow-up defects are worth stating because both were subtle:

- **The receipt-time fallback was still reaching the display.** `ParseLiveEvent` accepts a payload with no `timestamp` and falls back to `DateTime.UtcNow` so ordering and generated ids keep working. Assigning `RawTimestamp` straight into `_lastTradeAt` made it non-null in exactly that case, putting the process clock back under an observation label. `LiveDataEventModel` now carries `HasObservedTimestamp`, set from the `TryGetProperty` result, and only a genuine observation reaches the field.
- **Last-enumerated, not latest.** `UpdateSessionStats` runs while the response is enumerated raw; the `OrderBy(RawTimestamp)` happens afterwards in the display loop. A newest-first batch left the *oldest* trade showing.
- **Fixing the second alone would have split the pair.** Taking the maximum for the time while `_lastPrice` kept its unconditional last-enumerated assignment would show one trade's price beside another trade's time. Price and time are now set together in the same branch, so the row is always either *this price, observed then* or *this price, observed we don't know when*.

### 3 — One relabel

`TradeTimeText` is assigned `DateTime.Now` on refresh, so the *value* was honest and the *label* was not. Two sites called it a trade time; both now read **"Refreshed"**.

### Deliberately left alone

`OrderBookViewModel.LoadDemoTrades` is explicitly demo data, and `DataQualityViewModel` synthesizes x-axis labels for a sine-wave series. Every other `DateTime.Now` in `src/Meridian.Wpf/` is a legitimate refresh or audit stamp.

## Reason

Found while scoping `W10-MARK-001`. It is a small, self-contained correctness fix that does not need to wait for that row, and shipping a fund-operations blotter that answers "when was this observed?" with "just now, always" is not something to carry forward.

## Testing performed

- [x] Relevant unit tests were added — **six**, listed below
- [ ] `bash scripts/ci.sh` full quality-gate — **not run locally; `dotnet` is not installed in this environment**

`PositionBlotterViewModelTests`:

- `MapToEntry_ExecutionResponseCarriesNoObservationTime_LeavesMarketTimeUnknown` — the mapping produces `MarketTime == null` from a fully-populated response.
- `StatusBar_NoEntryCarriesAnObservationTime_RendersUnknownRatherThanAClockReading` — both tiles read `"—"`.

`LiveDataViewerViewModelTests`, all four driving events through session-stat accumulation in **response order** rather than sorted order, since that is where the defects lived:

- `LastTradeTime_ObservedTimestamps_ShowsTheLatestTradeNotTheLastEnumerated` — also asserts the price, so the pair cannot split.
- `LastTradeTime_PayloadOmittedTheTimestamp_StaysUnknownRatherThanShowingReceiptTime` — price still renders, time reads `--`.
- `LastTradeTime_BatchMixesObservedAndUnobserved_KeepsTheObservedTradeIntact`
- `LastTradeTime_SessionReset_ReturnsToUnknown`

`MapToEntry` widens from `private` to `internal`, and two `…ForTests` seams were added on the live-data view model, matching the existing `LoadEntriesForTests` pattern and the `InternalsVisibleTo("Meridian.Wpf.Tests")` already declared in `AssemblyInfo.cs`. Existing tests build `BlotterEntry` directly and set `MarketTime` themselves, so they are unaffected.

**Compilation and test results are verified by `verify-dotnet` on this PR, not locally.** Flagging that plainly rather than implying local verification that did not happen.

**Phase marker:** `PR9` (`src/**`, `tests/**`), confirmed via `tools/roadmap/enforce_phase_scope.py` read by **exit code**.

## Safety review

## Governance changes

---


rodoHasArrived:
Both findings confirmed against source and fixed in `e29983231`.

**Receipt time was still reaching the display.** `ParseLiveEvent` accepts a payload with no `timestamp` and falls back to `DateTime.UtcNow` (`LiveDataViewerViewModel.cs:472`) so ordering and generated ids keep working. Assigning `RawTimestamp` straight into `_lastTradeAt` made it non-null in exactly that case, putting the process clock back under a label that reads as a market observation — the defect this PR set out to remove. `LiveDataEventModel` now carries `HasObservedTimestamp`, set from the `TryGetProperty` result rather than re-derived, and only a genuine observation reaches `_lastTradeAt`.

**Last-enumerated, not latest.** Confirmed: `UpdateSessionStats` is called at `:378` while the response is enumerated raw; the `OrderBy(RawTimestamp)` is at `:384`, in the display loop that runs afterwards. A newest-first batch left the oldest trade in `_lastTradeAt`. It now takes the latest observed trade.

**One more, found while fixing the second.** Taking the maximum for the time while `_lastPrice` kept its unconditional last-enumerated assignment would have split the pair — "Last trade" price from one trade beside another trade's time. Price and time are now set together in the same branch, so the row is always either *this price, observed then* or *this price, observed we don't know when*. A payload with no timestamp can still supply a price while nothing better has been seen, but never displaces an observed trade.

Four regression tests over internal seams that feed events through accumulation in response order: latest-not-last selection, missing timestamp staying `--`, a mixed batch leaving the observed trade intact, and session reset returning to unknown. `dotnet` isn't available in this environment, so `verify-dotnet` is the only thing that compiles this — reporting whatever it returns.

---

PR #2561, opened 2026-08-01
Repair the 13 inert agent tool declarations

rodoHasArrived:
## Summary

Every agent under `.claude/agents/` declared:

```yaml
tools: ["read", "search", "edit", "mcp"]
```

**None of those four is a Claude Code tool name**, and the effect is not a reduced grant but an empty one: when nothing in `tools` resolves, the host refuses to launch the subagent and the Agent tool returns an unresolved-entry error. All **thirteen** agents were inert. I hit this directly when `meridian-blueprint` would not spawn and I had to fall back to the skill.

The vocabulary is traceable rather than a typo. `.claude/plugins/*/agents/*.md` carry VS Code chatmode names — `changes`, `codebase`, `edit/editFiles`, `runCommands`, `vscodeAPI`, `microsoft.docs.mcp` — and `.github/agents/software-engineer-agent-v1.agent.md` carries the same list. The Meridian files use an abbreviated form of it, so this looks like a copy from VS Code custom chat modes.

The host expects a **comma-separated string**, not a YAML sequence, with capitalised tool names: `tools: Read, Glob, Grep`.

> **The validator that was originally in this PR has moved to `claude/agent-definition-validator`.** It had absorbed most of seven review rounds' churn — a hand-rolled YAML subset parser generating fidelity finding after fidelity finding — and none of it was load-bearing for the repair. It is being rebuilt there on PyYAML. This PR is now the repair alone.

### Grants follow each agent's own stated workflow

| Agents | Grant | Why |
| --- | --- | --- |
| `-code-review`, `-repo-navigation` | `Read, Glob, Grep` | `code-review`'s own description says **"Findings only — no edits"** |
| `-simulated-user-panel` | `+ Agent` | its contract offers independent persona voices on explicit request; without `Agent` a requested panel collapses into a single voice while still calling itself a panel. No write access |
| `-blueprint`, `-brainstorm`, `-roadmap-strategist` | `+ Edit, Write, Bash` | they produce files, and AGENTS.md:131 makes `git status --short` a start gate before editing |
| `-archive-organizer`, `-browser-workstation`, `-cleanup`, `-docs`, `-implementation-assurance`, `-provider-builder`, `-test-writer` | `+ Edit, Write, Bash` | each mandates a command in its own workflow — a targeted test run, an `npm` script, a validation command, or a file move that `Edit`/`Write` cannot express since neither deletes the source |

### On `meridian-code-review` and git

A scoped grant — `Bash(git diff:*)`, `Bash(git show:*)` — was added mid-review so the agent could obtain a diff from a commit hash or branch, which its workflow step 1 assumed. It has been **reverted**, because Claude Code scoping is a command *prefix* match and therefore also admits `git diff --output=<file>`, which writes. The grant could not enforce the "Findings only — no edits" boundary the description advertises, and documenting the limitation left that boundary resting on an instruction rather than on the tool set.

Losing caller-free diff access is the accepted cost. The workflow now says so instead of assuming a diff is reachable: scope the review to named files, or to a diff the caller supplies.

### On the dropped `mcp` token

It never resolved. The host **does** accept `mcp__<server>` and `mcp__<server>__<tool>` patterns, so the intent is expressible — but which MCP servers exist is a property of the **host session**, not of this repository, and naming a concrete server in a checked-in file would break on hosts that lack it. The agents lose nothing they had; they had nothing.

## Reason

The agent layer was wholly non-functional. Thirteen broken files sat unnoticed because no check opened them — `validate-skill-packages.py` walks only `.claude/skills/`, and `check-ai-inventory.py` enumerates filenames without parsing frontmatter. The gate that closes that hole is the follow-up PR; this one restores the capability.

## Testing performed

- [x] All thirteen declarations are comma-separated strings with capitalised names — `grep -n '^tools:' .claude/agents/*.md`
- [x] Script-test lane after the validator removal — `459 tests, failures=0, errors=0`
- [x] `bash -n scripts/ci.sh`; `documentation.yml` parses; ci-workflow-contract and workflow-docs-parity suites green
- [x] `run-docs-automation.py --profile core` — no drift, no dangling validator reference

**Not verified:** that a repaired agent actually spawns. That needs a host session to exercise, and I would rather say so than imply I ran it.

**Phase marker:** `PR9` — `.claude/**` falls outside both `PR1` (`docs/**`) and `PR6`. Confirmed via `enforce_phase_scope.py` read by **exit code**.

## Safety review

## Governance changes

- [ ] `.github/workflows/**`
- [ ] `.github/CODEOWNERS`
- [ ] `AGENTS.md`
- [ ] `scripts/ci.sh`

The earlier revision of this PR **did** modify `.github/workflows/documentation.yml` and `scripts/ci.sh`, and was correctly flagged for governance review. Both changes belonged to the validator and have moved with it, so this PR no longer touches a protected file.

---


rodoHasArrived:
All fifteen findings applied in `4b272797c`. Nothing rebutted this round — every one held on inspection.

### Bash grants (5 findings)

The original pass scored each agent by searching for explicit command names, which returned zero for all thirteen files, so only two got read properly. Reading each documented workflow instead confirms all five:

| Agent | Step that needs a command |
| --- | --- |
| `meridian-test-writer` | 3 — "Run the targeted test project and report results" |
| `meridian-provider-builder` | 3 — "…run targeted tests" |
| `meridian-browser-workstation` | 3 — "Validate with targeted `npm --prefix src/Meridian.Ui/dashboard run test` or `run build`" |
| `meridian-cleanup` | 3 — "Run the narrowest relevant validation command" |
| `meridian-archive-organizer` | 3 — "Move/archive only with a reference trace"; `Edit` and `Write` can write the destination but cannot delete the source, so a *move* is not expressible without `Bash` |

The grant table in the description is superseded by this:

| Agents | Grant |
| --- | --- |
| `-code-review`, `-repo-navigation`, `-simulated-user-panel` | `Read, Glob, Grep` |
| `-blueprint`, `-brainstorm`, `-roadmap-strategist` | `+ Edit, Write` |
| `-archive-organizer`, `-browser-workstation`, `-cleanup`, `-docs`, `-implementation-assurance`, `-provider-builder`, `-test-writer` | `+ Edit, Write, Bash` |

### Validator (8 findings)

**Real YAML, duplicate keys rejected.** The frontmatter now parses once through a `SafeLoader` subclass whose mapping constructor raises on a repeated key. Both stated failures are covered by tests: a duplicate `tools` key (regex validates the first value, host resolves the last) and `description: >` with no body (parses to `''`; a regex saw the literal `>` as content). Field types are checked too, so a sequence is caught by *type* rather than by a leading `[`.

**Scoped syntax.** Entries split on top-level commas only, so `Agent(worker, researcher)` stays one entry and the head name before `(` is what gets validated. This form is not hypothetical here — `.claude/settings.local.json` already uses `Bash(dotnet:*)`, `Bash(git:*)`, and eight more.

**Field-sensitive `mcp__*`.** Accepted in `disallowedTools`, rejected in `tools`. Beyond what the host documents, the asymmetry stands on its own: a deny-list wildcard matching zero servers is harmless, while an allow-list wildcard matching zero servers *is* the empty grant this validator exists to catch.

**`KNOWN_TOOLS`.** All eight names you listed are in, alongside the rest of the current surface. There is no machine-readable host schema to source it from, so it is maintained by hand and **pinned by a test** — adding or removing a name fails until the change is made in both places, which is the "synchronize and cover with tests" half of your suggestion.

**`rglob`** for discovery, and **fail-closed**: a missing directory and an empty directory both return 1 rather than printing a note and passing.

### CI wiring (1 finding) — the one that made the rest reachable

Confirmed exactly as described. `documentation.yml:15-25` omitted `.claude/agents/**`, and `scripts/ci.sh` invokes `run-docs-automation.py --scripts <explicit list>` that excludes this validator, so an agent-only change met no hosted check. Both fixed: the path is added to the workflow's `pull_request` **and** `push` filters, and the validator now runs directly in `verify_docs` alongside its unit tests — so it is exercised by the always-run lane regardless of path filtering.

### Tests

`tests/scripts/test_validate_agent_definitions.py`, 33 cases: the tracked tree, the pinned vocabulary, the Bash-grant and findings-only postures, tool-list parsing, scoped entries and unbalanced parentheses, malformed / duplicate / non-mapping frontmatter, field-specific MCP patterns, and fail-closed discovery including a nested definition.

Locally: validator `13 file(s), 0 error(s)`; `33 tests … OK`; `bash -n scripts/ci.sh` clean; ci-workflow-contract and workflow-docs-parity suites still green; full regeneration sequence rerun after merging `main`, which had also left this branch `dirty`.

---

rodoHasArrived:
CI went red on `4b272797c` — my fault, one root cause across all three lanes. Fixed in `e68579d3e`.

```
File "build/scripts/docs/validate-agent-definitions.py", line 53, in <module>
    import yaml
ModuleNotFoundError: No module named 'yaml'
```

Taking the "parse it as real YAML" finding literally, I imported PyYAML at module scope. **PyYAML is optional in this repository by design** — `build/scripts/docs/common.py:207-216` guards its import behind `try`/`except` and falls back to `_parse_yaml_subset`, and the only pin is `tools/source_docs/requirements-render.txt`, which the hosted Python lanes do not install. So the import took out `verify-docs` directly, `verify-workflows` through the new test module (`errors=1` in the script-test lane), and `regenerate-docs` through the automation profile.

The fix parses the frontmatter subset directly instead: plain scalars, folded and literal block scalars, flow sequences, block sequences. Every property the round asked for survives — duplicate keys rejected rather than silently last-wins, invalid lines reported rather than skipped, a bodyless `description: >` reading as empty, sequences distinguished by type — and there is now **one code path in every environment**, which is what a gate needs. A validator that behaves differently depending on whether an optional dependency happens to be present is worse than either behaviour alone.

PyYAML still earns its keep, as a **differential oracle in the tests**: where it is installed, the suite asserts the hand-rolled parser agrees with `yaml.safe_load` on every tracked definition and that both reject the same malformed input. Those two cases skip where it is absent, so the lane is green either way and the parser stays honest wherever anyone can check it.

Verified with PyYAML blocked by a meta-path hook, reproducing the CI environment:

- validator — `13 file(s), 0 error(s)`, exit 0
- new suite — `Ran 40 tests … OK (skipped=2)`, the two skips being the differential pair
- full script-test lane (with PyYAML, as locally available) — `499 tests, failures=0, errors=0`
- `run-docs-automation.py --profile core` — `validate-agent-definitions → success`, no drift

---

rodoHasArrived:
Four more findings, all applied in `09b0e2ca0` (three code) and in the PR description (governance). I reproduced each of the three code findings against the committed validator before fixing — **all three failed open**, which is the worst direction for a check whose whole job is catching an inert agent layer.

```
comma-only tools         -> NO ERRORS (fail-open)
misspelled tool key      -> NO ERRORS (fail-open)
unterminated quote       -> NO ERRORS (fail-open)
```

**Punctuation-only tool list.** `tools: ","` is non-empty, so the emptiness check cleared it, then split to zero entries and granted nothing — the exact empty-grant state this validator exists to prevent, waved through by the validator itself. A split yielding no entries is now an error pointing at the same remedy as the empty case.

**Misspelled permission field — the most serious of the four.** The asymmetry is what makes it dangerous: omitting `tools` *inherits the default pool*, so `tool:` instead of `tools:` does not narrow a grant, it silently widens it. A deliberately findings-only agent would come back with edit and command access and nothing in the repository would notice. Frontmatter keys are now validated against the supported set (`name`, `description`, `tools`, `disallowedTools`, `model`, `color` — the tracked tree uses only the first three), with a near-miss check that names the actual consequence per field:

> unknown frontmatter key `tool`; did you mean `tools`? an omitted `tools` inherits the default tool pool, so this misspelling widens the grant instead of narrowing it

> unknown frontmatter key `disalowedTools`; did you mean `disallowedTools`? an omitted `disallowedTools` denies nothing, so this misspelling silently drops the restriction

An unrelated unknown key is still rejected, without a misleading hint.

**Unterminated quoted scalar.** `description: "unterminated` survived because the unquote helper returned it as a plain string. A real parser fails the whole document, so the host could not load the agent at all — a green result on a definition that cannot exist. It now raises. Good catch on flagging this as fresh evidence *after* the hand-rolled parser landed; it was introduced by that change, not pre-existing.

**Governance.** Correct, and I should have caught it when I wrote the step. `.github/workflows/**` and `scripts/ci.sh` are both in AGENTS.md's Protected Governance Files list, and the description declared none changed. The description is rewritten: both files are now checked with a line each on what changed and why, the PR is explicitly designated for human governance review, and there is a banner at the top so it is not missed. Both edits are additive — no existing step, trigger, or filter entry removed or reordered — and they exist precisely because the validator is otherwise unreachable for the class of change it guards, which is the finding that prompted them.

Six tests added (46 total). Verified with PyYAML blocked: `46 tests, 0 failures, 2 skipped`. Full script-test lane: `505 tests, failures=0, errors=0`. Validator on the tree: `13 file(s), 0 error(s)`.

**On the previous run:** `quality-gate` went red on `e68579d3e`, but the three lanes the PyYAML fix targeted all passed — `verify-docs`, `verify-workflows`, and `verify-browser` were `success`. The gate failed solely because `verify-dotnet` was `cancelled`, superseded by this push.

---

rodoHasArrived:
Both applied in `a7c94ac77`. Both reproduced against the committed parser first.

### `description: null` — fail-open, and my differential test should have caught it

```
description: null    parsed={'description': ('str', 'null')}   errors=(clean)
description: ~       parsed={'description': ('str', '~')}      errors=(clean)
```
```python
>>> yaml.safe_load('description: null\ntools: 123\n')
{'description': None, 'tools': 123}
```

Confirmed exactly as described: the parser stored the literal string, `required_string` accepted it as non-empty, and the file validated clean — while the host resolves the field to null and the agent loses the description it routes on. A definition that passes the gate and does not work.

Plain scalars now resolve with YAML semantics matching PyYAML (so YAML 1.1 — `yes`/`no`/`on`/`off` are booleans). That fixes the reported case and a second one it implies: `tools: 123` and `tools: true` are now reported by type rather than silently becoming one-entry tool lists. Quoted scalars skip resolution, so `description: "null"` stays a genuine string. Exotic YAML 1.1 numeric spellings (hex, octal, underscores, sexagesimal) are deliberately left as strings, documented in the function — none appears in agent frontmatter.

Worth noting why this got past the differential test I added last round: **it only compared the tracked definitions**, none of which uses a null scalar, so the oracle never saw the disagreeing input. That was the actual defect — a differential test whose fixtures are only the happy path. It now runs explicit fixtures across null, tilde, bare-key, capitalised, int, float, bool, `yes`/`off`, quoted, folded, and block-sequence forms, and asserts the **resolved type** rather than just the value. That is the check that would have caught this unprompted.

### `KNOWN_FIELDS` was too narrow — and this one is on me

Correct, and the consequence is worse than a normal false positive because I made this a **required** gate in the same PR. A definition using `permissionMode: plan` would have failed CI, so the check I added to prevent permission *widening* would have blocked permission *hardening*. The allowlist now covers the documented host surface — `permissionMode`, `skills`, `hooks`, `memory`, `background`, `isolation` alongside the original six — with a comment saying why completeness is load-bearing rather than tidiness.

The near-miss check that justified rejecting unknown keys is unaffected: none of the added names is within one edit of `tools` or `disallowedTools`, so `tool:` still fails with its hint. Verified both directions:

```
skills/hooks/etc     -> (clean)
still catches typo   -> unknown frontmatter key `tool`; did you mean `tools`? …widens the grant…
```

Four tests added (50 total). Verified with PyYAML blocked — `50 tests, 0 failures, 3 skipped`. Full script-test lane — `509 tests, failures=0, errors=0`. Validator on the tree — `13 file(s), 0 error(s)`.

---

rodoHasArrived:
All six applied in `3b09b2631`. Three of them were bugs I introduced last round by widening the field allowlist without widening anything else — worth naming plainly, because the pattern is the point.

### Nested `hooks` — I broke this by adding the key

```
nested hooks -> ['frontmatter is not valid YAML: unexpected indented content on line 5']
```

Adding `hooks` to `KNOWN_FIELDS` while the parser accepted only scalars and `- ` items under a key meant the gate rejected a documented hook block it had just declared supported. That is the second time widening the allowlist alone created a new failure — the first was `permissionMode` being blocked, this is the field being accepted but its value being unparseable.

The parser is now genuinely recursive: mappings, sequences, and block scalars nest to any depth, with duplicate-key detection at every level, and a key with an inline value that also has an indented child still errors. Nested fixtures were added to the PyYAML differential so this class stays covered rather than covered-by-luck.

### Field values — applied, but deliberately partial, and I want to be explicit about why

`permissionMode: plna` and `background: maybe` both passed. Values are now checked per field.

But I have **not** value-checked every optional field, and that is a considered choice rather than an omission. Restricting a field to a value set that turns out to be incomplete blocks legitimate work — which is precisely how my first `KNOWN_FIELDS` failed two rounds ago, and value domains are a strictly easier thing to get wrong than key names. So:

- `permissionMode` → validated against `default`, `acceptEdits`, `bypassPermissions`, `plan`
- everything else → **shape**-checked (`background` boolean, `hooks` mapping, `skills` string-or-sequence, the rest strings)

Shape checks catch the failure you described — a mapping written as a scalar, a boolean written as prose — without my inventing a vocabulary I cannot verify from here. If you have the authoritative sets for `isolation` or `memory`, I will add them; I would rather say "not validated, here is why" than guess a second time.

### MCP-only grants

Confirmed, and it contradicted the validator's own docstring, which already says MCP servers are a host-session property. `tools: mcp__github__*` alone resolves to nothing on a session without that server — the empty grant that made the layer inert. An allow-list now requires at least one built-in. Deny-lists are unaffected: one that matches nothing is harmless, which is the same asymmetry behind the `mcp__*` field-sensitivity rule.

### Agent grants — both verified against source

**`git status --short` is genuinely mandated.** AGENTS.md:131, step 1 of the Agent Workflow: *"Run `git status --short` before editing and treat unrelated changes as user-owned."* So `blueprint`, `brainstorm`, and `roadmap-strategist` could not perform their own start gate. Granted.

**`code-review` step 1 is diff-scoped:** *"Scope the diff or target files."* `Read`/`Glob`/`Grep` cannot produce a diff from a commit hash or branch, so a normal commit review was unusable. Granted **scoped**, not blanket — `Bash(git diff:*)`, `Bash(git show:*)`, `Bash(git log:*)`, `Bash(git status:*)` — which keeps the "Findings only — no edits" posture the description promises. A test pins that it never holds unscoped `Bash` and that every `Bash(...)` entry it does hold stays git-scoped, so this cannot quietly widen later. The scoped-syntax support added earlier in this PR is what makes that expressible.

`repo-navigation` and `simulated-user-panel` are unchanged — their workflows are orientation and manifest-driven review, which `Read`/`Glob`/`Grep` do cover.

### Docs

Both maintained references updated: `.github/workflows/README.md:41` now records the `.claude/agents/**` trigger and the new validation, and `build/scripts/docs/README.md` catalogs the validator with its fail-closed rules and local reproduction commands.

Nine tests added (59 total). PyYAML blocked: `59 tests, 0 failures, 3 skipped`. Full script-test lane: `518 tests, failures=0, errors=0`. Validator on the tree: `13 file(s), 0 error(s)`.

---

rodoHasArrived:
All four applied in `9d1c7b190`. Reproduced first, as before:

```
unterminated flow    -> (clean - FAIL OPEN)
hook mapping entry   -> ['unexpected indented content on line 7']
allow cancelled      -> (clean - FAIL OPEN)
```

### Two of these came from the fix before them, and that pattern is now the story of this PR

Last round I widened the parser to accept nested mappings. That fix handled `key:` followed by an indented mapping — but left `_parse_sequence` treating every `- ...` item as a scalar, so `- matcher: Bash` with sibling keys beneath it, **the ordinary shape of a hook entry**, still failed. I fixed the case I tested and not the case the field actually takes. Compact block mappings in sequence entries now parse, and the result matches `yaml.safe_load` exactly on a realistic `PreToolUse` block:

```
ours  : {'hooks': {'PreToolUse': [{'matcher': 'Bash', 'hooks': [{'type': 'command', 'command': 'echo hi'}]}]}}
pyyaml: {'hooks': {'PreToolUse': [{'matcher': 'Bash', 'hooks': [{'type': 'command', 'command': 'echo hi'}]}]}}
```

That fixture is now in the differential, along with flow-sequence cases — which is the actual lesson. Every round where I widened what the parser accepts, I opened a new way for it to accept something the host rejects, and each time the gap was in a shape I hadn't written a fixture for. The differential only helps where it has inputs.

**Unterminated flow sequence** is the same class: `skills: [foo` had its bracket stripped and returned `['foo']`. The closing delimiter is now required.

### The deny-list cancellation is the one I like most

`tools: Read` beside `disallowedTools: Read` passed because the two fields were checked independently — each valid on its own, and together an empty grant. It reaches the exact failure state this gate exists to prevent by a route **neither field-level check could see**, which is a nice demonstration that per-field validation is not the same as validating the definition. The effective sets are now compared, matching a scoped entry by head name so `disallowedTools: Bash` correctly cancels `Bash(git diff:*)`.

### The git grant — I can't fix this the way you asked, so I've said so instead

You are right that `Bash(git diff:*)` permits `git diff HEAD^ HEAD --output=review.md`, and that this defeats the "Findings only — no edits" boundary the description advertises. I verified the option exists on both `git diff` and `git show`.

But Claude Code's scoping is a **command prefix match**, so `Bash(git diff:*)` cannot express "except `--output`", and `disallowedTools` matches tool names rather than argument patterns. There is no way to write the restriction you asked for in the grant syntax. That leaves three options: drop the grant and re-break diff-based review, keep it and quietly overstate the boundary, or keep it and describe it accurately.

I took the third. The agent file now carries an explicit **Tool boundary** section stating that the git grants are read-only *by instruction*, that `--output`/`-o`/redirection are forbidden, and that the prefix scoping cannot mechanically exclude them — so the no-edit rule is enforced by the instruction rather than by the grant alone. An overstated boundary is worse than a documented one, and this PR has spent several rounds correcting exactly that kind of claim.

If you would rather have the mechanical guarantee at the cost of the capability, say so and I will drop the git grants and note that diff-based review needs the caller to supply the diff.

Six tests added (64 total). PyYAML blocked: `64 tests, 0 failures, 3 skipped`. Script-test lane: `523 tests, failures=0, errors=0`. Validator on the tree: `13 file(s), 0 error(s)`.

---

rodoHasArrived:
All four applied in `03fedb0da`. Reproduced first:

```
bad escape       -> (clean - FAIL OPEN)
mcp survives     -> (clean - FAIL OPEN)
bare dash hook   -> ['line 6 is not a `key: value` entry']
dontAsk          -> ["'dontAsk' is not a known mode; expected one of acceptEdits, bypassPermissions, default, plan"]
```

**`dontAsk` is the one I said would happen.** Two rounds ago I wrote that value-checking a domain I cannot verify from here risks blocking legitimate work, and that this was the failure mode of my first `KNOWN_FIELDS`. Then I added `PERMISSION_MODES` anyway and it was incomplete. Added, pinned test updated. The broader position is unchanged and now better evidenced: `permissionMode` is the **only** field with a value check, and every other optional field stays shape-checked. If you have authoritative sets for `isolation` or `memory` I will add them; I am not going to guess a third time.

**Deny-list leaving only MCP entries** — this is the sharpest of the four. `tools: Read, mcp__github__*` with `disallowedTools: Read` passed *both* rules I added last round: a built-in was declared, so the MCP-only check was satisfied; and not every entry was cancelled, so the cancellation check was satisfied. The empty grant only appears in the **intersection**, which neither rule computed. The MCP-only rule is now re-applied to the surviving set rather than to the declared one.

**Bare dash** — my compact-mapping fix handled `- key: value` but not `-` alone with the mapping indented beneath. That's the third consecutive round where a parser widening left an adjacent form still broken. Both the sequence loop and the sequence-versus-mapping dispatch now recognise it, and the differential confirms it:

```
ours  : {'hooks': {'PreToolUse': [{'matcher': 'Bash', 'hooks': [{'type': 'command'}]}]}}
pyyaml: {'hooks': {'PreToolUse': [{'matcher': 'Bash', 'hooks': [{'type': 'command'}]}]}}
equal : True
```

**Escapes** — correct that matched quotes are not well-formedness. Double-quoted scalars are now checked against YAML escape rules including the sized `\x`, `\u`, `\U` forms; single-quoted scalars are deliberately untouched, since YAML does not process escapes in them, and a test pins that distinction so the fix does not overreach.

Six tests added (70 total). PyYAML blocked: `70 tests, 0 failures, 3 skipped`. Script-test lane: `529 tests, failures=0, errors=0`. Validator on the tree: `13 file(s), 0 error(s)`.

---

rodoHasArrived:
All three applied in `d9c3e54a1`. Each is the same shape as the defect this PR exists to fix — a declaration promising something the runtime does not deliver — which is worth naming, because two of them were introduced by *my* fixes in the last two commits.

## Navigation MCP — confirmed, and the allow-list was the wrong shape

`.claude/skills/meridian-repo-navigation/SKILL.md:23-30` is explicit:

> If MCP access is available, prefer the generated navigation resources/tools before broad codebase searching:
> `mdc://repo-navigation/quick-start` · `mdc://repo-navigation/catalog` · `find-subsystem` · `route-task` · `find-entrypoints` · `find-related-projects` · `find-authoritative-docs`

An allow-list of `Read, Glob, Grep` suppresses every one, so the skill's own routing instruction becomes unfollowable and the agent silently degrades to filesystem search.

They also cannot simply be allow-listed instead. I checked: there is **no `.mcp.json` and no server definition anywhere in the repository** — the navigation server is host-session provided. Naming a concrete server in a checked-in file is exactly what this PR's own reasoning rejects, since it breaks on hosts that lack it.

So `meridian-repo-navigation` now declares `disallowedTools: Edit, Write, NotebookEdit, Bash`. The read-only posture is preserved *mechanically*, and a session that does provide navigation tooling can be used as intended. The file carries a **Tool boundary** section explaining why it is the one agent using a deny-list.

I have **not** switched `meridian-code-review` to the same shape, though the case is arguable — its skill's step 1 is an MCP PR fetch, which would restore diff access without reintroducing the `--output` hole. You chose the allow-list for a mechanical guarantee two rounds ago and an allow-list is additionally fail-closed against tools the host adds later, so I am not reversing that without your say-so. Say the word if you want it.

## Persona nesting — you are right, and it makes the grant possibly inert

This is the one that matters most, because it means the option chosen last round may not achieve its goal. I cannot read the host docs from here, but the corroborating evidence in this very session is strong: Claude Code's own first-party `Explore` and `Plan` subagents are both documented as **"All tools except `Agent`…"**. Nesting is at minimum not the intended pattern.

I have kept the `Agent` declaration — it is harmless if unusable and correct if usable — but the file no longer treats it as a guarantee. It now says to fall back to voicing personas in-session **and to name that fallback**, and records that genuinely independent voices require the *parent* session to launch the workers and pass results in for synthesis. The note ends on the point I think is the real one:

> a panel that silently collapses to one voice while still labelling itself a panel is the same class of defect as an agent whose declared tools do not resolve. Both look like they work.

If you would rather remove the independent-persona mode outright — the option you declined last round, before this limitation was on the table — that is a one-line change and I will make it.

## AI index sync

Correct, and required by AGENTS.md:139. `docs/ai/agents/README.md` now records all three contract changes in its Claude agent table: that `code-review` needs a caller-supplied diff, that `repo-navigation` uses session MCP via a deny-list, and where persona orchestration actually has to happen.

## Note on the branch

This PR is now the **repairs alone**; the validator moved to #2564, which is stacked on this branch and must merge after it. The three changes above are merged into #2564 too, so its suite stays aligned with the definitions it asserts against — including a rewritten `test_findings_only_agents_cannot_write` that checks the *posture* rather than assuming every findings-only agent uses an allow-list.

Script-test lane on this branch: `459 tests, failures=0, errors=0`.

**On the previous run:** `quality-gate` went red on `2f2e702c5`, but the run's own conclusion is `cancelled` — superseded by this push. That is the ninth such notification in this session; the check-run object reports `failure` while the workflow run reports `cancelled`, so only the run is authoritative.

---

PR #2562, opened 2026-08-01
Serve the prometheus-net registry at /metrics and prove the alerts fire (PRD-111)

rodoHasArrived:
## Summary

`/metrics` never served the metrics the alerts named. `StatusEndpointHandlers.GetPrometheusMetrics` hand-wrote ten legacy series into a `StringBuilder` and stopped, never touching the `prometheus-net` default registry — so the 81 `Metrics.Create*` declarations in `PrometheusMetrics.cs`, `WriteAheadLog.cs`, and the Lean integration types were incremented at runtime and never scraped. Ten of the eleven alerts named a series neither served by that handler nor declared in the registry. Only `MeridianDown`, on Prometheus's own `up{job="meridian"}`, could fire.

**The exporter.** `GetPrometheusMetricsAsync` appends `Metrics.DefaultRegistry` after the legacy block. Anything already scraping the ten original series sees byte-identical leading content; the declared instruments become reachable. Both call sites — `StatusHttpServer` and the `Ui.Shared` minimal-API mapping — are async and pass cancellation. `PrometheusExporterTests` pins the legacy names, a registry-declared probe, real `PrometheusMetrics` instrumentation, and the append ordering.

**The contract.** All 11 alerts, both Grafana panels, and all 8 `SloDefinitionRegistry` objectives are repointed onto served series. All 11 `runbook_url` anchors resolve to the active [Operator Runbook](docs/operators/operator-runbook.md), which replaces an archive stub, alongside active [Service Level Objectives](docs/operators/service-level-objectives.md) carrying SLIs, targets, error budgets, and RTO 2h / RPO 1h matching the thresholds `invoke-production-recovery.ps1` enforces.

## Reason

`PRD-111` in `docs/product/implementation-todo-list.md`. The row asked for executable service objectives and incident operations; the prerequisite nobody had noticed was that the exposition endpoint served none of the signals involved. This was split out of #2556 precisely because repointing alerts without fixing the exporter reaches the same permanent silence by a different route.

Three gates hold it, and each was mutation-tested rather than assumed:

| Gate | What it catches | Mutation verified |
| --- | --- | --- |
| `validate-observability-contract.py` (33 tests) | unexported metric, unresolvable runbook/SLO anchor, alert with no `AlertRunbookEntry`, literal admin credentials, **and an exposition endpoint that stops serving the registry** | removing the `CollectAndExportAsTextAsync` call fails the gate |
| `promtool test rules` against `deploy/monitoring/alert-rules.test.yml` | a rule that parses but does not fire on its stated condition | reverting the WAL alert to `rate(...)` fails the tests |
| `docker compose config` | base stack that will not start, or a monitoring overlay with a default admin credential | `${GF_SECURITY_ADMIN_PASSWORD:-admin}` fails the gate |

The exporter check matters most: without it, every other check in the contract gate was reasoning about "emitted" metrics while the endpoint emitted none of them, so the gate would have reported a clean contract over alerts that could not fire.

`promtool check rules` only proves the file parses. Every alert defect this repository has hit was of the second kind — a `rate()` window over a counter advanced once at startup that cleared a P1 integrity alert while the corruption stood, an equality test that went silent during half-open recovery, a `for:` paired with a mismatched range, thresholds on gauges with no writer that fire permanently on a healthy instance. The unit tests pin firing behaviour, and the negative cases matter as much as the positive ones.

`validate-monitoring-deployment.py` fails closed when either tool is missing: an unrun check reports the same green as a passed one. CI installs promtool; local runs pass `--allow-missing-tools`.

The monitoring services live in a separate `docker-compose.monitoring.yml` because Compose interpolates the merged file before it filters services by profile — a `${VAR:?}` in the base file broke `docker compose up` for **every** profile, not just monitoring.

## Testing performed

- [ ] `bash scripts/ci.sh` completed successfully — not run end to end; the lanes exercised are listed below

| Lane | Result |
| --- | --- |
| `dotnet test tests/Meridian.Tests -c Release --filter PrometheusExporter\|SloDefinitionRegistry\|AlertRunbookRegistry` | 28 passed |
| `promtool check rules deploy/monitoring/alert-rules.yml` | SUCCESS, 11 rules |
| `promtool test rules deploy/monitoring/alert-rules.test.yml` | SUCCESS |
| `validate-monitoring-deployment.py --summary` | 3 checks passed; 3 mutations each correctly rejected |
| `validate-observability-contract.py --summary` | 79 emitted metrics, 11 alerts, 8 SLOs, 0 findings |
| `tests/scripts/test_validate_observability_contract.py` | 33 passed |
| `check-workflow-hygiene.py` | passed |
| `run-docs-automation.py --profile core` + diagram/structure/manifest regeneration | clean, artifacts committed |

## Known gaps, stated rather than hidden

`MeridianDataFreshnessViolation` and `MeridianSlaComplianceLow` stay guarded (`and ... > 0`) and honestly silent. Nothing calls `PrometheusMetrics.UpdateSlaMetrics` or `RecordSlaFreshness`, so those gauges sit at their default 0 and an unguarded threshold would page permanently on a healthy instance. A promtool test fails if a guard is removed without `DataFreshnessSlaMonitor` being wired into the periodic exporter. The same applies to `mdc_processing_latency_microseconds`, whose only caller is a test — the latency alert uses `mdc_average_latency_microseconds`, which has a real writer.

`PRD-111` therefore stays open. What remains is a dated alert, incident, diagnostics-bundle, and recovery-drill artifact: real operational execution, not code.

## Safety review

- [x] No secrets or credentials were committed — the change removes the `admin`/`admin` Grafana default and adds a gate against its return

## Governance changes

- [x] `.github/workflows/**` — one `Install promtool` step added to the `verify-docs` job
- [x] `scripts/ci.sh` — two validation steps added to the docs lane

Governance-file changes require explicit human approval. Both exist to run the two tools this PR's correctness argument depends on; without them the gates are present but never execute in CI.

---
_Generated by [Claude Code](https://claude.ai/code/session_01Kz5UBbuEf499B7afwga2T6)_


rodoHasArrived:
Pushed `52224072d`. All six findings confirmed against source before acting; two were mine and serious enough that this PR was not mergeable as it stood.

**The exposition was invalid, not merely incomplete.** The legacy block emitted `mdc_events_per_second` and `mdc_historical_bars_per_second`, and `PrometheusMetrics` registers both families too. Concatenating produced a second `HELP` line for the same name. promtool on the concatenated output:

```
error while linting: text format parsing error in line 10:
second HELP line for metric name "mdc_events_per_second"
```

Prometheus rejects the **entire scrape** on a parse error, so once `PrometheusMetrics` initialised — `EventPipeline.RecoverAsync` → `RecordWalRecovery` does it — `/metrics` served nothing at all. My change was strictly worse than the defect it set out to fix. Both families are removed from the legacy block; the registry owns them under the same names, so nothing scraping them sees a rename. `GetPrometheusMetricsAsync_DeclaresEachMetricFamilyExactlyOnce` parses every `HELP`/`TYPE` line and asserts uniqueness — the `Contains`/`StartsWith` assertions I had could not see this, which is exactly why it got through.

**The registry was served but never written.** `PrometheusMetricsUpdater` is constructed only in `PrometheusMetricsUpdaterTests` — no host or DI call site — and `UpdateFromSnapshot()` is called only from inside that unused updater. Every gauge it feeds sat at its initial zero, so the drop-rate, latency, circuit-breaker, and validation alerts would have been evaluating zeros. Served-but-unwritten reads identically to a healthy system, which is the same class of defect as declared-but-unserved, one layer down.

`GetPrometheusMetricsAsync` now refreshes on the scrape rather than wiring a background service: the values are current exactly when read, and there is no new timer or lifecycle to own. That also makes removing the two legacy duplicates safe — the registry copies now carry live values, not zeros.

**Remaining four:**

| Finding | Change |
| --- | --- |
| A green `promtool test rules` only runs the cases present | `check_every_alert_has_a_firing_test` compares deployed alert names against tested ones and fails on omissions, with an explicit `DISABLED_ALERTS` exemption. It found the four you named. Added those scenarios plus three silence cases; the gate now passes on its own terms. |
| `MeridianNoEventsPublished` cold-start suppression | Guarding on a prior publication permanently suppressed a process whose provider is broken from the first second — the exact outage. Now guarded on the series' own age (`offset 10m`): silent at 5m, fires at 15m, both tested. |
| Market-hours suppression | **Not fixed.** No session signal is exported, so the rule still pages during closed sessions once the window is real. Stated in the rule rather than left implicit; needs a session gauge first. |
| `StatusHttpServer` cancellation | Passes `_cts.Token` to collection and to the response write. My previous commit message claimed this and was wrong. |
| SLO doc's RTO/RPO | `invoke-production-recovery.ps1` computes `measuredRpoSeconds` as `backupCompleted - operationStarted` and `measuredRtoSeconds` as `restored - backupCompleted` — backup and restore durations, excluding detection, decision, reconciliation, and acceptance, and timing a backup taken *during* the drill so it cannot express data loss at all. The table now marks both objectives stated-not-measured and says what would close them. |

Validation: 6 exporter tests, 18 with `PrometheusMetrics`; `promtool check rules` 11 rules, `promtool test rules` SUCCESS; `validate-monitoring-deployment.py` 3 checks; `validate-observability-contract.py` 0 findings; `dotnet format` clean solution-wide.

---

rodoHasArrived:
## `verify-dotnet` on `ffca69e5c` — flake, not this PR

One test failed: `MaintenanceSchedulerTests.ScheduledJob_WhenRunningAdmissionCannotBeRetained_IsObservedAndRequeued`, on a `WaitUntilAsync` poll with a five-second budget. Everything else passed — 1319, 336, 1658, 994, 433, 1710, 177, 3144, 134, 432, 1251, 350, 10, 65, 104, 3, 17, 29.

Ran it three times locally: 19/19 passed each time. None of this PR's 30 changed files touches storage, the scheduler, or anything they depend on — `git diff --name-only origin/main...HEAD | grep -i storage` is empty. It's a timing-sensitive wait under runner load. Not re-pushing to chase it; it should clear on the next run.

## The four new review findings

Two need verification I have not finished, and two are real:

**`/metrics` behind `LoginSessionMiddleware`** — if the middleware runs before the mapping and does not exempt `/metrics`, the scrape gets a redirect or 503 and *nothing* in this PR reaches Prometheus, including `up{job="meridian"}`. That would subsume every other finding here. I have not yet confirmed the middleware ordering or its exempt-path set, and I am not going to assert either way without doing so.

**Resettable totals mirrored into monotonic counters** — this one is a direct consequence of my refresh-on-scrape change. `BackfillCoordinator.RunAsync` calls `Reset()` on the same static snapshot the scrape now reads; `IncTo` cannot decrease, so `mdc_events_published_total` and `mdc_events_dropped_total` would flatten after a backfill and `MeridianNoEventsPublished` could fire during an active feed. Serious, and mine.

**`MeridianNoEventsPublished` is global, SLO-DC-002 is per-symbol** — correct. One live stream masks any number of stalled symbols. The rule does not cover the objective it is linked to.

**No Alertmanager** — correct. `docker-compose.monitoring.yml` provisions Prometheus and Grafana only, and `prometheus.yml` configures no receiver, so nothing pages anyone. The SLO document should say "visible in Prometheus" rather than "pages a responder".

## Where this leaves the PR

This is the fifth review round, and the pattern is consistent: the monitoring stack was never wired end to end at any layer, and each fix exposes the next unwired one — declared but not served, then served but not written, then written from a source nothing mutates, and now possibly not reachable by the scraper at all.

I have stopped adding scope and am waiting on a decision. What this PR honestly delivers today is a working exporter, an exposition that parses, eleven alerts each proven to fire by promtool, and three mutation-tested CI gates. What it does not deliver is live signal behind several of those alerts. Landing it with the gaps documented and opening a follow-up for the writer wiring, the middleware exemption, and Alertmanager is the option I would pick; continuing to grow it is the alternative.

---

PR #2566, opened 2026-08-04
fix(security-master): write factor schedules and swap legs in the shape their readers accept

rodoHasArrived:
## Summary

Fixes the two Security Master term shapes that were written narrower than they were read, and adds the architecture review that identified them (`docs/architecture/security-master-architecture-review.md`).

Both defects were type-correct at the top level while their contents disagreed with every consumer, so a security could be created through the canonical path, pass every validator, and still be economically unusable — silently.

**Factor schedule.** `StructuredCreditTerms.FactorSchedule` was `string option`, serialized as free text, and declared `String` in `SecurityAssetTermsSchema`. But `StructuredCashFlowTermsResolver.ReadFactorSchedule` and `SecurityMasterAccountingEventSourceAdapter` both accept only an array. CLO/CMO/MBS tranches therefore restated face off the scalar `currentFactor` for their whole life, while `RequiresFactorSchedule` (set when `currentFactor &lt; 1`) raised a High-severity break that blocked paydown accounting on a condition the write path could not satisfy. It is now a typed `FactorScheduleEntry list`.

**Swap legs.** F# `SwapLeg` carried 4 fields against `StructuredCashFlowLeg`'s 11 — no notional, no payment frequency, no day count, so no computable payment. It now carries `legId`, `direction`, `spreadBps`, `currentIndexRate`, `notional`, `paymentFrequency`, `dayCount`, and `exchangesPrincipal`.

The schedule also flows into the economic-terms `structuredProduct` block as `{asOfDate, priorFactor, currentFactor}` transitions, because that is where the accounting-event adapter actually reads it — fixing only the asset-terms side would have left the paydown coverage gate unsatisfiable.

Backward compatible: legs in the pre-widening shape still deserialize, and a legacy free-text `factorSchedule` is preserved in a new `factorScheduleNote` rather than dropped on re-write.

Regression guard: `SecurityAssetTermField` can now declare per-element shapes for `Array`/`Object` fields, declared for swap legs, factor schedules, principal schedules, and covenants. Leaving nested shapes undeclared is exactly how both defects escaped the existing codec tests.

## Reason

Scheduled architecture review of the Security Master against institutional finance requirements. The review found the platform's classification, capability, validation, and identifier lanes to be genuinely data-driven and test-defended, with risk concentrated in one structural pattern: write-side domain records narrower than read-side typed models, bridged by tolerant alias-probing that yields nulls instead of errors. This PR closes that gap for the two shapes where it affects accrual and paydown accounting. The review document records the remaining findings (R2–R8) for separate follow-up.

## Testing performed

- [x] GitHub Actions `quality-gate` passed

Added `SecurityMasterTermScheduleCodecTests` (11 cases) which round-trip C# → F# domain → serialized JSON and then hand the result to the real read-side resolver, so the two sides are measured against each other rather than each against its own assumptions. Added 3 F# cases in `DomainTests.fs` covering serialization and domain validation.

The authoring environment had no .NET SDK, so the first push was unbuilt; the `verify-dotnet` failure it produced (a missed `StructuredProductTerms` construction site, `FS0764`) is fixed in `205b3162`. CI remains the authoritative gate.

`verify-dotnet` was green across all 19 .NET test projects at `6121b2b5`, including the three lanes that carry this change: `core-execution-strategy` (`Meridian.Tests.SecurityMaster`, 1873 passed), `core-platform-domain-root` (`FactorPaydownProjectionServiceTests`, 3155 passed), and `fsharp` (`DomainTests.fs`, 439 passed — up from 436 by the three cases added here). `135bc6cb` merges `origin/main` and re-runs that verification against the merged tree.

## Known unrelated failures on this PR

- **`schema-control`** — reproduced locally against a clean `origin/main` checkout, which produces an identical stale-artifact list (ledger, reporting, catalog manifests, and the per-schema manifests). This branch added exactly two entries to that list, both caused by the `ElementFields` property; both are regenerated and committed here. The remaining drift is pre-existing and left alone rather than swept into this diff. Independently corroborated: `schema-control.yml` runs on `main` only by manual dispatch, and its most recent `main` run (`dc39eecd`, 2026-07-27) also failed.

Two checks previously listed here have since been resolved and are now green:

- **`regenerate-docs`** — was mine after all, not a `main` failure. Running `schema-control` locally left an untracked `build/schema-control/candidate/` tree; the doc-health and repository-structure generators walk the filesystem rather than the git index, so the committed dashboards were generated against polluted counts. Corrected in `c8ed9d25`.
- **`verify-desktop`** — flaky, not caused by this change. Confirmed two ways at the same commit with no code change: the manual `Targeted Test` workflow ran `Meridian.Wpf.Tests` on `windows-latest` in Release with a *superset* of the CI filter (Integration cases included) and passed, and a re-run of the failed job itself then went green. Nothing under `src/Meridian.Wpf/` or `tests/Meridian.Wpf.Tests/` references any type or catalog this PR touches.

## Deferred follow-ups

Recorded in the review document and in [this comment](https://github.com/rodoHasArrived/Meridian-main/pull/2566#issuecomment-5190422929); each needs a product decision rather than a mechanical fix, and two are downstream consumer bugs this PR merely made reachable:

1. Governed profile version bump — `SecurityAssetProfileCatalog:70` still declares `factorSchedule` as `RequiredText`.
2. Widening `SecurityEconomicTermsV2ToAssetSpecificTermsUpcaster` to carry `structuredProduct`.
3. Scalar-vs-schedule authority for `structuredProduct.factor`.
4. `BuildLegSchedule` opening principal exchange.
5. Cash-flow projector treating `FactorAsOf == 0` as missing and defaulting face to 100%.

## Safety review

## Governance changes


rodoHasArrived:
Thanks — I verified each of the twelve comments against source before acting. All were legitimate; ten are fixed in `fd83f500`, two are deferred with reasoning below.

**The headline: paydown accounting was still blocked.** Three findings compounded, and together they meant the central claim of this PR was not yet true.

`FactorPaydownProjectionService` validates the evidence list *before* its equal-factor `NoChange` short-circuit (`Project`, lines 69–88). My economic-terms rows carried `asOfDate`/`priorFactor`/`currentFactor` only, so `SecurityMasterAccountingEventService` passed `[]` as evidence and every paydown was rejected as `factor-paydown.evidence-required` — exactly as `Generate_FactorPaydownWithoutEvidence_ShouldFailClosed` covers. `FactorScheduleEntry` and `FactorSchedulePoint` now carry `Source`/`EvidenceLink`/`SourceContentHash` through to economic terms.

That ordering also refines the no-change point: retaining unchanged observations is only safe *because* evidence now flows — a no-change row without evidence would still be `Rejected`, not `NoChange`. Both fixes were needed together.

**Fixed:**

| Finding | Resolution |
|---|---|
| Evidence missing from schedule rows | `Source`/`EvidenceLink`/`SourceContentHash` carried end-to-end |
| Unchanged observations dropped | All observations emitted; period coverage preserved |
| Increasing factors accepted | Rejected at the domain boundary (`structured_credit_factor_schedule_not_monotonic`) |
| `SpreadBps >= 0` rejected SOFR − 10bp | Constraint removed — negative spreads are ordinary economics |
| `legType` aliases declared but unread | `rateType`/`type` honoured on the write path |
| `exchangesPrincipal` aliases unread | `principalExchange`/`notionalExchange` honoured |
| `Payer`/`Receiver` rejected | Normalized to `Pay`/`Receive` on read; unrecognized tokens pass through so the domain still reports them |
| Duplicate factor dates silently collapsed | De-duplication removed so the domain validator actually sees them |
| Codec test checked names, not types | Now asserts `ValueKind` against each declared field type |
| Dashboard TODO count 220 vs 221 | Regenerated; both now agree |

**Deferred — these need a product decision, not a mechanical fix:**

1. **Profile-backed structured credit** (`SecurityAssetProfileCatalog:70` declares `factorSchedule` as `RequiredText`). Correct, and it means operators creating MBS/CLO records through the approved profile still submit a scalar. Changing it is a governed profile **version bump** through `SecurityAssetProfileGovernanceService`, affecting existing records that reference the approved version — not something to slip into this PR unreviewed.

2. **`SecurityEconomicTermsV2ToAssetSpecificTermsUpcaster` does not carry `structuredProduct`.** Correct: economic-only records (no `LegacyAssetSpecificTerms`) lose the schedule on rebuild. This is pre-existing — the upcaster never carried `structuredProduct` at all — but this change makes it newly load-bearing, and widening it touches every asset class that rebuilds through that path.

Both are worth doing; I'd rather they land deliberately than as a silent widening here. Note also that nothing in this PR has been built or tested locally — this environment has no .NET SDK, so CI remains the authority.

---

rodoHasArrived:
Verified all three against source. Two are the same defect class this PR exists to close and are fixed in `85918767`; the third is a read-model change and joins the deferred list.

**Fixed — the container probe was narrower than the reader.** I routed the row *fields* through `SecurityTermReader` in `abf26fe0` but left the *container* probe on `JsonElement.TryGetProperty`, which is case-sensitive, while `SecurityTermReader.TryGetProperty` matches `OrdinalIgnoreCase`. A payload spelling its container `FactorSchedule` was therefore readable by `StructuredCashFlowTermsResolver` and round-tripped to an empty schedule by the write path — exactly the asymmetry this change is about, one level up from where I fixed it. `GetOptionalArrayItems` and the legacy-note probe now use the shared reader, so `legs`, `principalSchedule`, and `covenants` get the same treatment.

**Fixed — a wrong-typed container was indistinguishable from an absent one.** `ValueKind != Array` fell through to `yield break`, so a vendor sending a single object instead of an array persisted an empty list while the create reported success. `ReadFactorScheduleContainer` now throws for a present-but-wrong-typed container, while keeping the three cases that must stay silent: absent, empty (so the alternate spelling is still probed), and the pre-typed free-text string, which `GetFactorScheduleNote` still carries forward.

Both are covered by new cases in `SecurityMasterTermScheduleCodecTests`.

**Deferred — `BuildFactorHistory` only probes `structuredProduct.factorHistory`.** Correct, and the consequence you describe is right: accounting consumes the schedule while the operator-facing factor timeline stays empty. This is the same category as the two read-model items already deferred on this PR — `SecurityEconomicTermsV2ToAssetSpecificTermsUpcaster` and the governed profile version bump. Emitting `factorHistory` as a second alias would be the cheap fix, but it duplicates the schedule in the economic-terms payload under two keys with no single authority, which is the pattern that produced the scalar-vs-schedule ambiguity in the first place. Teaching the read model the new shape is the right fix and belongs with the other read-model work, developed against a compiler rather than through CI round-trips.

Standing caveat: this environment has no .NET SDK, so nothing here is locally built. CI remains the authority. `quality-gate` and `verify-dotnet` were green on `6339298c` immediately before these edits.

---

rodoHasArrived:
Triage status — main has moved under this PR in a way a rebase can't reconcile mechanically, so it stays open for a decision rather than being auto-refreshed like the rest of the queue:

- **Factor schedules:** main now models typed schedules its own way — `FactorScheduleEntries: FactorScheduleEntry list` (with the legacy `FactorSchedule: string option` retained) parsed via `ToFactorScheduleEntry` from a `factorScheduleEntries` key. This PR's writer-side `factorSchedule` array (asOfDate/priorFactor/currentFactor + evidence keys) is a competing shape for the same problem. Note main's `SecurityEconomicDefinitionAdapter` still emits **no** schedule rows at all, so the reader-shape gap this PR described may persist under the new model — worth re-verifying against `SecurityMasterAccountingEventSourceAdapter` as it reads today.
- **Swap legs:** the alias-tolerant, widened `ToSwapLeg` here (legId/id/name, rateType/type aliases, optional per-leg economics) does not appear on main and may still be wanted as-is.

Recommendation: re-cut as a smaller PR carrying (1) the swap-leg widening and (2) whatever adapter emission gap survives under `FactorScheduleEntries`, then close this one. It sits 7 real source conflicts away from main and its green CI predates the model change (2026-08-05).

---

PR #2568, opened 2026-08-05
Provision the agent validator's dependency and correct its permission matching

rodoHasArrived:
## Summary

Follow-up to #2564, which merged with four review findings open. Four behaviour changes:

- **The `core` profile's dependency is now provisioned everywhere it runs.** `production-certification.yml` invokes `run-docs-automation.py --profile core`, that profile includes `validate-agent-definitions`, and the file installed nothing at all. Both `--profile core` callers now install the pinned requirements. `ci.yml` passes an explicit `--scripts` list that excludes the validator, so it needs none. `scripts/ci.sh` gets a preflight that names the install command rather than pip-installing into a developer's own interpreter.
- **Scoped-permission containment is corrected.** Comparing raw frontmatter strings meant `"git push:*".startswith("git:")` was `False`, so a broad deny did not cancel a narrow allow and the validator reported an effectively empty grant as surviving — the defect class it exists to catch, in itself. Scopes normalise to a command prefix, which also reconciles the two wildcard spellings in use: the `command:*` form Claude Code writes on permanent approval and the documented `command *` glob. The word boundary is preserved, so `Bash(git:*)` covers `git push` but not `gitfoo`.
- **Partial MCP tool wildcards are accepted.** `mcp__github__get_*` is documented as matching that server's `get_` tools and the pattern rejected it. The bare `mcp__server` form is deliberately kept — it resolves to every tool on that server, so rejecting it would block working declarations for no safety gain.
- **Two omissions the review did not reach.** `KNOWN_FIELDS` was short by four documented fields, not two — `maxTurns`, `mcpServers`, `effort`, `initialPrompt` — and `PERMISSION_MODES` was missing `auto` and `manual`, the same omission that produced the earlier `dontAsk` failure. Both are now pinned by tests asserting a literal set, since a test that iterates the constant passes no matter what the constant leaves out.

## Reason

#2564 shipped with these open. Every claim here was checked against the Claude Code reference documentation rather than accepted from the review or written from memory, which changed three of the four outcomes:

| Review claim | Verified outcome |
| --- | --- |
| `production-certification.yml` runs the validator unprovisioned | **Confirmed** — no `pip install` anywhere in the file |
| Broad deny fails to cancel narrow allow | **Confirmed** — and the fix needed the documented word-boundary rule, not a bare `startswith` |
| `mcp__server__tool` is required; bare `mcp__server` invalid | **Half rebutted** — bare server entries are valid and are kept; only the partial-wildcard rejection was a real defect |
| `maxTurns` and `mcpServers` missing | **Confirmed, and understated** — four fields were missing, and `PERMISSION_MODES` was short by two |

One thing I could not reproduce: whether the unprovisioned lane actually fails on the runner. The only recent `documentation-evidence` job that passed without an install ran on a branch whose head contains zero occurrences of `validate-agent-definitions`, so it predates the merge and proves nothing. No certification run has executed with the validator registered. The fix stands on the design point — depending on an undeclared transitive dependency is the fragility this work set out to remove — not on an observed failure.

## Testing performed

- [x] `bash scripts/ci.sh` completed successfully — targeted lanes run; `dotnet` is unavailable in this environment, so the .NET lanes are CI-only
- [x] Relevant unit tests were added or updated — 81 validator tests, including eleven containment cases covering both wildcard spellings, both directions, and the `gitfoo` boundary
- [ ] Relevant integration tests were added or updated — not applicable; no integration surface changes

Also run: validator 13 files / 0 errors, full script-test lane 658 tests / 0 failures / 0 errors, `run-docs-automation.py --profile core` exit 0, `test_ci_workflow_contract.py` and `test_check_workflow_docs_parity.py` green, `bash -n scripts/ci.sh` clean. The preflight was exercised against an interpreter with `yaml` blocked and exits 1 with the install command.

## Safety review

## Governance changes

- [ ] No governance files changed
- [x] `.github/workflows/**` — one step added to `production-certification.yml`'s `documentation-evidence` job
- [ ] `.github/CODEOWNERS`
- [ ] `AGENTS.md`
- [x] `scripts/ci.sh` — one preflight step added to `verify_docs`

Governance-file changes require explicit human approval. Both are additive single steps; neither changes an existing command.


rodoHasArrived:

rodoHasArrived:
All three applied in `b98e77660`, each confirmed against the permission reference. Two are defects in the containment function this PR itself introduced — the fix I wrote for the original scope finding was too narrow in two independent ways.

**Wildcards may appear anywhere.** Confirmed: *"Wildcards can appear at any position in the command, including at the beginning, middle, or end"*, with `Bash(git * main)` and `Bash(* install)` as the worked examples. My `_scope_prefix` recognised only a trailing `*`, so a mid-string or leading deny fell through to the exact-match branch and cancelled nothing — reintroducing exactly the effectively-empty-grant blind spot the original finding was about, one shape over. Scope coverage is now a **glob match** rather than a prefix test.

**`:*` is a command alias only on the shell tools.** Confirmed, and this is the sharper of the two. `Tool(param:value)` is a parameter match on any tool, and *"the value supports `*` as a wildcard… so `Agent(isolation:*)` matches any explicit isolation value"*. Stripping that colon as a command separator turned `WebFetch(domain:*)` and `WebFetch(domain:example.com)` into the prefixes `domain` and `domain:example.com`, and the bounded check then returned false — the deny matched nothing.

The alias is now confined to `Bash` and `PowerShell`, and the reference gives the principled reason rather than it being a special case: parameter matching *cannot* target a tool's primary content field, and that field is exactly `command` for those two. So `param:` can never be a parameter prefix there, which is what makes the alias unambiguous where it applies.

**The attribution was backwards.** Also confirmed — *"The permission dialog writes the space-separated form when you select 'Yes, don't ask again' for a command prefix. The `:*` form is only recognized at the end of a pattern."* I had it the other way round in the README and in the code comment. Corrected, and the "only at the end" half is now a test: *"In a pattern like `Bash(git:* push)`, the colon is treated as a literal character"*, so that grant survives a `Bash(git:*)` deny.

Thirteen containment cases now cover both directions, both spellings, leading/middle/trailing wildcards, the parameter form on two non-shell tools, a different-parameter deny that must **not** cancel, the `gitfoo` word boundary, and the colon-literal case. Validator 13 files / 0 errors, 86 validator tests, 663 script-lane tests, docs automation core profile clean, every changed path exits 0 against `enforce_phase_scope.py --phase PR9`.

Separately: the merge conflict was already resolved in `a1c85e2` before the `@copilot` request landed — the two conflicting files were generated dashboards, resolved to `main`'s side and regenerated from the merged tree.

---

rodoHasArrived:
All three applied in `05fc2c013`. These reviewed `0fcd402681`, two commits back, so I checked each against current code rather than assuming — one was partly fixed already, and it turned out to have a second shape I had missed.

**Non-suffix Bash wildcards.** `Bash(git * main)` was already handled by `b98e77660`. But the example I hadn't tried, `Bash(* --help *)`, still cancelled nothing: the trailing-` *` branch escaped its **whole prefix** literally, so a wildcard *inside* the prefix matched an asterisk character rather than acting as a glob. That is the same correction `b98e77660` made for the non-bounded branch, applied one branch over — I fixed half the function and left the other half with the original bug. Both branches now share one `_glob_to_regex` helper, so they cannot drift again.

**MCP tool globs.** Confirmed. The tool segment is a glob, so `mcp__github__*_issue` and `mcp__github__get_*_issue` are valid and were being rejected as unknown tools. Again the same too-narrow reading — accepting only a trailing star. The **server** segment stays glob-free, since an allow rule must name a specific configured server and an unanchored glob there is skipped with a warning rather than honoured; `mcp__*` remains deny-only.

**`mcpServers` and the MCP-only guard.** Confirmed, and this one is an interaction between two changes in this same PR rather than a pre-existing gap. The guard rejects an allow-list of only MCP entries because server availability is a property of the host session — but adding `mcpServers` to the recognised fields makes the servers a property of the *definition*, so `tools: mcp__playwright` alongside an `mcpServers` entry now resolves and was being rejected. I added the field without noticing it invalidated the guard's premise. The guard is now conditional on `mcpServers` being absent, and both directions are tested.

Verified: 91 validator tests, 668 script-lane tests, validator 13 files / 0 errors, docs automation core profile clean, every changed path exits 0 against `enforce_phase_scope.py --phase PR9`. Regression checks held — the narrow-deny, `gitfoo` word-boundary, and colon-literal cases still behave correctly.

Worth saying plainly: this is the third consecutive round where reviewing this permission-matching code found real defects in it, and twice now the defect was in a fix I had just written. The rules have more corners than I've been able to enumerate up front, so further findings here should be expected rather than treated as surprising.

---

rodoHasArrived:
All three applied in `0241696b0`. Every one is a defect in code added earlier in this PR, and two of them are consequences of the `mcpServers` exemption I added in the *previous* commit.

**WebFetch domain scopes are not a plain glob.** Confirmed, and this is the one with a security-shaped rationale rather than just a correctness one. The reference: *"In any position other than a leading `*.` or a bare `*`, the wildcard matches only the text between two dots. `WebFetch(domain:example.*)` matches `example.org`… but not `example.evil.com`, where `*` would have to become `evil.com` and cross a dot. This keeps a trailing wildcard from matching domains an attacker could register."*

Generalising to a glob was the right move for Bash and the wrong one here — it made a deny cancel a grant the host leaves live, i.e. reporting an empty tool set that isn't empty. Domain scopes now use domain matching: a leading `*.` spans subdomains at any depth but excludes the bare domain, a bare `*` matches everything, any other wildcard is confined to one label, and comparison is case-insensitive with a trailing `.` stripped from both sides. Seven cases pinned, including the `example.evil.com` non-match and `*.example.com` not covering `example.com`.

**The `mcpServers` exemption was blanket, not per server.** Confirmed — I wrote `not frontmatter.get("mcpServers")`, so *any* truthy value suppressed the guard for *every* entry. `mcpServers: [playwright]` with `tools: mcp__github` passed, though a session without github resolves nothing. That's the same empty-grant hole the guard exists to close, reopened by the fix for the previous round's finding. The exemption is now keyed on the declared-server set.

**`mcpServers` members were unvalidated.** Confirmed, and it compounds the one above: `- 123` was accepted *and* counted as a declaration that suppressed the guard. Entries are now required to be a server name or a keyed inline definition; malformed members are reported and contribute nothing to the declared set.

Verified: 96 validator tests, 673 script-lane tests, validator 13 files / 0 errors, docs automation clean, phase gate clean on every changed path. Earlier regressions all still hold — narrow-deny, `gitfoo` boundary, colon-literal, internal wildcards, and the MCP glob positions.

This is the fourth round on this code, and the pattern is consistent enough to name: each fix generalises correctly for the case in front of it and over-generalises somewhere adjacent. The Bash glob was right for Bash and wrong for WebFetch; the `mcpServers` exemption was right in principle and too broad in application. I've stopped treating "the tests pass" as evidence the semantics are complete here — the tests only cover cases I thought to write.

---

rodoHasArrived:
All five applied in `5eaa25dd2`. Each verified against the references rather than accepted on description — which mattered for two of them.

**Normalisation ran on one side only.** Confirmed. I normalised the deny and then matched its regex against the *raw* allow text, so `disallowedTools: Bash(git *)` did not cover `tools: Bash(git:*)`. My earlier test happened to use `Bash(git push:*)`, which starts with `git ` and so matched by accident — the direction that actually exercised the gap was the one I didn't write.

**Parameter scopes ignore colon whitespace.** Confirmed — *"Whitespace around the colon is ignored"*. One wrinkle worth recording for anyone writing such a rule: `disallowedTools: Agent(model: *)` must be **quoted** in YAML frontmatter, because unquoted `model: *` parses as a nested mapping rather than a scalar. My first attempt at the test failed on exactly that, which is a real trap for definition authors, not just for the test.

**PowerShell is case-insensitive.** Confirmed — *"Matching is case-insensitive"*, and PowerShell "uses the same shape as Bash rules". Only PowerShell relaxes; Bash stays case-sensitive, and there's a test pinning both directions. I deliberately did **not** implement the alias canonicalisation the same paragraph describes (`PowerShell(Get-ChildItem *)` also matching `gci`, `ls`, `dir`). Guessing at the host's alias table would produce false *cancellations* — reporting a live grant as dead — which is the more damaging error for a check whose whole job is deciding whether a grant survives. It's noted in the code as a known limitation.

**`mcpServers` must be a sequence.** Confirmed by the worked example, which settles what the prose alone left ambiguous:

```yaml
mcpServers:
  - playwright:
      type: stdio
  - github
```

A top-level mapping is the `.mcp.json` shape, not the frontmatter one. Accepting it let a definition the host cannot use suppress the MCP-only empty-grant guard — precisely the thing that exemption must never do.

**The exemption didn't survive deny filtering.** Confirmed, and it's the same mistake as the "one side only" one above: I added the declared-server exemption to the per-field guard and not to the survivor check, so `tools: Read, mcp__playwright` with `mcpServers: [playwright]` and `disallowedTools: Read` failed at the second gate. An exemption has to follow its rule to every place the rule is enforced.

**Three of my existing tests were wrong** and are corrected: two encoded the top-level mapping as the valid shape, and one used an unquoted colon-space scope that wasn't valid YAML at all. They failed when the code got stricter, which is what they were for.

103 validator tests, 680 script-lane tests, validator 13 files / 0 errors, phase gate clean.

Two rounds ago I said further findings here should be expected rather than surprising. That's held: rounds four and five each found real defects, and in both the recurring shape was *applying a correct rule to only some of the places it belongs* — one side of a comparison, one of two gates. That's a more specific failure mode than "the semantics are subtle", and it's the one I'd look for first in any further round.

---

rodoHasArrived:
All four applied in `c1a8f8f4`. Three of them are the same shape of mistake, which is worth naming.

**@607 — WebFetch deny scopes weren't normalized.** Confirmed and fixed. `_scope_covers` normalized `allow_scope` and then tested the **raw** `deny_scope` for the `domain:` prefix, so `WebFetch(domain : example.*)` missed the domain branch entirely, fell through to the generic glob — whose `.*` crosses a dot — and cancelled a grant for `example.evil.com` that the host leaves live. Both sides now normalize before the branch is chosen.

This is the third round running that a finding has taken the form of *applying a correct rule to only some of the places it belongs*: normalization on the deny but not the allow, the exemption on the field guard but not the survivor check, and now the branch selector reading a raw value the rest of the function had already normalized. I said last round that was the first thing I'd check in any further round, and it recurred anyway — so the fix here is placed at the single point where the branch is chosen rather than at each call site.

**@772 — MCP denies were compared for string equality.** Confirmed, and it was worse than a missed cancellation. `disallowedTools: mcp__github` didn't cancel `tools: mcp__github__get_issue`; the survivor was then MCP-only with its server declared, so the declared-server exemption suppressed the *second* guard too. A fully cancelled grant passed both gates clean. New `_head_covers` treats MCP names as hierarchical — a bare server covers its tools, `mcp__server__*` covers the matching subset, `mcp__*` covers everything — while preserving direction the way scopes already do: a per-tool deny narrows a whole-server grant rather than erasing it, and a wildcard on the allow side is compared literally rather than expanded.

**@531 — inline `mcpServers` values were unvalidated.** Confirmed. `- slack: nope` declared `slack`, and that name then bought an MCP-only grant an exemption the host would not honour, since without a server config the grant resolves to nothing. The value must now be a config mapping. The exemption is only as sound as the declaration it trusts. The two malformed shapes report distinctly — a non-mapping value points at `drop the colon to reference an already configured server by name`, which is the actual fix and is not obvious from a generic "malformed entry".

**@474 — PowerShell aliases.** I declined this last round; on reconsideration that was the wrong call for the part of it that is knowable. `gci`, `ls`, and `dir` are named explicitly in the reference, so canonicalizing exactly those three strictly increases detection with no chance of being wrong about them. What I was actually right about was the rest of the table, and that limitation stands: it isn't enumerable from the docs, so an undocumented alias is still compared literally. That under-reports rather than over-reports, which is the correct direction to fail for a check that decides whether a grant survives. Canonicalization rewrites the command token only — `PowerShell(ls*)` keeps its glob, because it also covers `lsof` and is not the alias — and `Bash` gets none of it, since `ls` and `dir` are unrelated commands in a POSIX shell.

**Verification.** Twelve tests added. I checked them against the previous commit rather than asserting they cover the findings: **nine fail** on `5eaa25dd2` and pass here. The three that pass on both are direction-preservation guards, which is what they're for.

```
validate-agent-definitions.py         13 file(s), 0 error(s)   exit 0
test_validate_agent_definitions.py    115 tests, OK
run-script-tests.py                   692 tests, failures=0, errors=0
run-docs-automation.py --profile core exit 0
check-ai-inventory.py --summary       322 items, 0 findings
enforce_phase_scope.py --phase PR9    exit 0 on all 7 changed paths
```

The only regenerated content is the doc-health line count (120,364 → 120,383, from the README edit); orphaned files unchanged at 255.

---

rodoHasArrived:
Correction to the verification block above: the phase gate covers **5** changed paths, not 7, and I posted that line before running it rather than after. Run now, by exit code, on each path in `c1a8f8f44`:

```
0 build/scripts/docs/README.md
0 build/scripts/docs/validate-agent-definitions.py
0 docs/status/doc-health-dashboard.json
0 docs/status/doc-health-dashboard.md
0 tests/scripts/test_validate_agent_definitions.py
```

The result stands — the gate passes — but a stated count that doesn't match the diff is exactly the kind of thing that makes an evidence block worth less than no evidence block, so it shouldn't sit uncorrected. Every other figure in that comment was observed before it was written.

The four review comments that just landed are against `5eaa25dd2e`, the commit before the fixes; they're the same four findings addressed in `c1a8f8f44` and in the reply above, so no further change is needed for them.

---

rodoHasArrived:
Two more rounds landed. The findings against `5eaa25dd2e` and `c1a8f8f448` that re-arrived are already fixed — `c1a8f8f44` and `a6b2b33c0` respectively, both covered in the replies above. Two were genuinely new.

**@456 (`a6b2b33c0`) — `mcp__github__*` must cancel a bare `mcp__github`.** Correct, and it's my own fix from the round before overreaching. I wrote "a per-tool deny leaves the rest of the server standing" and implemented it as `return False` for *any* deny carrying a tool segment. But `mcp__github__*` isn't a per-tool deny — it names every tool the server has, so against a bare-server grant it covers it exactly. Now only a tool segment of nothing but wildcards counts, and `mcp__github__get_*` against `mcp__github` is pinned as still surviving so the boundary can't drift back.

**@608 (`a6b2b33c0`) — an inline config that's a mapping but not a usable one.** Also correct for the case it names. `isinstance(config, dict)` accepted `{}`, which cannot reach a server, so it exempted an MCP-only grant that resolves to nothing. The config must now declare a transport: `command` for stdio, `url` for http/sse.

I checked the reference rather than taking the whole finding: every server example at `code.claude.com/docs/en/mcp` carries one or the other, and none carries neither — so requiring a transport can't reject a config that works. **I did not act on the `url`-only half.** All 13 remote examples pair `url` with `type`, but the reference never states `type` is required, and inferring http from `url` is plausible host behaviour. Rejecting `{url: ...}` on a 13-for-13 correlation would be guessing at a schema, and that's the mistake that made the first `KNOWN_FIELDS` block legitimate agents. `{url: ...}` without `type` is accepted, with a test pinning it.

That's the second time in this PR that verifying a finding against primary docs split it into a real half and an over-tightening half. It keeps being worth the fetch.

**Verification** — all figures observed before writing them this time:

```
validate-agent-definitions.py         13 file(s), 0 error(s)   exit 0
test_validate_agent_definitions.py    119 tests, OK
run-script-tests.py                   696 tests, failures=0, errors=0
run-docs-automation.py --profile core exit 0
enforce_phase_scope.py --phase PR9    exit 0 on all 5 changed paths
```

Five new tests across the two fixes; four fail against their preceding commit, and the fifth (`{url: ...}` accepted) is the boundary guard that must pass on both.

---

rodoHasArrived:
Two of the three applied in `c4b9bceda`. The third I implemented, measured, and backed out — detail below.

**@724 — Read/Edit globs crossing path segments.** Confirmed, and documented verbatim: *"In gitignore patterns, `*` matches within a single path segment and can appear at any position in the pattern, while `**` matches across directories."* So `Read(src/*.json)` does not cancel `Read(src/a/b.json)`, and the Bash-style `.*` said it did. This is the WebFetch domain bug again on a different tool — a generic matcher applied where the tool has its own scope grammar. Fixed with a segment-aware translator for `Read`/`Edit`/`Write`.

I deliberately stopped short of full gitignore semantics. The reference also specifies anchoring (`//`, `~/`, leading `/`), bare filenames matching at any depth, and a depth difference between allow and deny rules for single-segment directory patterns — all of which resolve against *the settings source that defined the rule*, which an agent file doesn't have. Modelling them from a subagent definition would be guessing. The omission under-reports genuine cancellations, which is the safe direction, and it's stated in both the code and the README rather than left implicit.

**@567 — Bash wrapper stripping.** Confirmed, and this one deserves an explicit retraction. When I declined the PowerShell alias canonicalisation two rounds ago, my stated reason was that the host's table isn't enumerable from the docs and guessing creates false cancellations. That reasoning was right for PowerShell — but I then carried the assumption to shell normalisation generally without checking. The Bash wrapper list **is** published in full, and the reference says it is *"built in and ... not configurable"*: `timeout`, `time`, `nice`, `nohup`, `stdbuf`, `command`, `builtin`, `noglob`, bare `xargs`, plus leading environment assignments. Nothing to guess.

Reproduced with the documented exclusions, each of which would otherwise invent a cancellation: `command -v` looks a command up rather than running one; `nocorrect` is absent from the list; `xargs` is stripped **only** with no flags (`xargs -n1 grep` is matched as an `xargs` command); and environment runners — `devbox run`, `npx`, `docker exec` — are explicitly excluded. Applied to the **command** side only: the deny is a rule, and a rule reading `Bash(timeout 30 npm test)` matches nothing, because every command reaching it has already been stripped. Env assignments are stripped unconditionally because this comparison only ever runs deny-against-allow, and *"a deny or ask rule matches past any leading assignment"* — the narrower known-safe-variable rule governs allow matching, which never happens here.

**@614 — string `mcpServers` entries. Implemented, then reverted.** The semantic claim is correct and I verified it: the reference says an entry is *"either a server name referencing an already-configured server ... or an inline definition with the server name as key and a full MCP server config as value."* So a string reference genuinely is weaker evidence than an inline definition.

But the conclusion doesn't follow. I built it — string references exempt only when the repo's own `.mcp.json` configures the server — and it broke **five existing tests**, every one of them the documented `mcpServers: [name]` shape. This repo has no `.mcp.json`, so the effect was to make the primary documented form unusable. A string reference isn't provably empty the way `{}` as an inline config is: it works whenever the user has that server configured, which is the normal case for anyone running the agent. Rejecting it is the same over-tightening that made the original `KNOWN_FIELDS` block legitimate agents — the exact failure this PR exists to correct.

So the exemption stays as-is for string entries. Recording the attempt rather than a bare "won't fix", since the distinction is real and someone may want it behind a repo that *does* ship `.mcp.json`.

**Verification:**

```
validate-agent-definitions.py         13 file(s), 0 error(s)   exit 0
test_validate_agent_definitions.py    127 tests, OK
run-script-tests.py                   704 tests, failures=0, errors=0
run-docs-automation.py --profile core exit 0
enforce_phase_scope.py --phase PR9    exit 0 on all 5 changed paths
```

Thirteen tests added; all thirteen fail against `c78f6d42d`.

---

rodoHasArrived:
All four applied in `8b46d9f16`. One of them is a correction to my own last commit.

**@160 — `Write` path rules are never consulted, and adding `Write` was my error.** The reference is explicit: *"Claude Code checks file permissions against `Edit(path)` and `Read(path)` rules only. If you write a path rule for `Write`, `NotebookEdit`, `Glob`, or the legacy `MultiEdit` tool instead, Claude Code accepts the rule but never consults it, and warns at startup."* So `Write(src/*.json)` cancels nothing — the host never reads it.

The finding named Read and Edit; **I added `Write` on my own initiative** last round, and shipped a test asserting the wrong behaviour along with it. That test is now corrected rather than deleted, and the real rule is pinned in both directions: a *scoped* deny on `Write`/`NotebookEdit`/`Glob` cancels nothing, while the *unscoped* form still does, because *"Claude Code doesn't warn about a tool-name rule with no path, such as a deny rule for `Write`; it matches that rule at the tool level everywhere."*

Widening a documented rule past what the docs support is the same failure as the over-tightenings earlier in this PR, just in the other direction — and this one I did to myself while fixing a finding about being too generic.

**@134 — deny-side tool-name globs.** Confirmed: *"Deny and ask rules also accept glob patterns in the tool-name position. The pattern must match the full tool name."* `disallowedTools: mcp__github*` was rejected as unknown; it's now accepted and applied, along with `B*` and `*`. The allow side keeps its restriction, and the docs give the reason: *"An unanchored allow glob such as `"*"`, `"B*"`, or `"mcp__*"` is skipped with a warning and doesn't auto-approve anything"* — which makes it an empty grant, not a broad one, so `tools: B*` stays an error. Typo detection is untouched, because acceptance is keyed on the `*`: `disallowedTools: Raed` still fails.

Handling the server-segment glob generally also let `mcp__*` stop being a hard-coded special case — it's now just the common spelling of a server glob.

**@777 — `WebFetch(domain:*)` cancels a bare `WebFetch`.** Confirmed: *"`WebFetch(domain:*)` matches every domain and is equivalent to a bare `WebFetch` rule."* The domain semantics were already right; the direction guard was skipping scoped denies against unscoped grants before they could be consulted. A scope covering everything the tool can do is now treated as equal to no scope, with `WebFetch(domain:example.com)` still only narrowing.

**The PyYAML gate.** Reproduced with an import shim: `run-script-tests.py` returns `errors=1` without PyYAML, from this module's import. But I fixed it in the test module rather than in `scripts/ci.sh`, for two reasons:

1. The validator exiting 2 is right for a CLI and wrong to inherit in a test module — it takes out a run of ~700 unrelated tests over one optional dependency. It now converts to a module-level skip naming the install command. Same shim, after: `578 tests, OK, skipped=37`.
2. A `verify_workflows` preflight isn't needed. `.github/workflows/meridian-ci.yml:186-187` installs `build/scripts/docs/requirements.txt` in that job before calling the lane, so these tests always run in CI — the skip only reaches an unprovisioned local shell, which is exactly where a clear message beats a stack trace. That also avoids a second change to a protected governance file for a case CI already covers.

**Verification:**

```
validate-agent-definitions.py         13 file(s), 0 error(s)   exit 0
test_validate_agent_definitions.py    134 tests, OK
run-script-tests.py                   711 tests, failures=0, errors=0
run-script-tests.py (no PyYAML)       578 tests, OK, skipped=37   ← was errors=1
run-docs-automation.py --profile core exit 0
enforce_phase_scope.py --phase PR9    exit 0 on all 5 changed paths
```

---

PR #2582, opened 2026-08-08
Repair main: restore the build and clear five red CI lanes

rodoHasArrived:
## Summary

**`main` has not compiled since 2026-08-06.** Six compile errors, all left by `ffeb200a2` ("robustness improvements") and the `codex/deepen-broker-statement-connectors` merges around it.

It stayed invisible because `verify-dotnet` runs `dotnet format` *before* the build step — the format gate failed first, so no compiler ever ran and no compiler error ever reached a log. `verify-desktop` and CodeQL were failing on the same broken build, reported only as "WPF development validation failed" and "configuration error".

I installed the SDK at CI's pinned version (10.0.302) to diagnose this, so everything below is locally verified rather than inferred.

## The compile errors

Every one is the same shape: **a merge kept a call site and took the declaration from the other side.**

| File | Error |
| --- | --- |
| `AlpacaBrokerageGateway.cs:63/69` | duplicate `AccountActivityPageSize` — two branches each added it, the merge kept both |
| `AlpacaBrokerageGateway.cs:666` | `activityResult.Cursor` — the paged `AlpacaActivityFetchResult` fetch was dropped, its call site retained |
| `StatementImportService.cs:620-626` | `SanitizeArtifactValue` called seven times, definition gone |
| `ConnectionStatusWebhookTests.cs:160` | `_messages` became a `ConcurrentQueue`; one `.Add` call site left behind |
| `PaperSessionRecoveryConcurrencyTests.cs:611/625` | `OrderStatus` ambiguous — both declaring namespaces imported |
| `LiveTradingEngineTests.cs:124/622/623` | `PaperTradingGateway` ambiguous; `StrategyRunStatus` unimported |

**Judgement calls, made on evidence:**

- **Alpaca cursor** — dropped the argument rather than restoring the paged fetch. `AlpacaActivityFetchResult` is entirely gone from `main`, and **nothing in `src/` reads `BrokerageActivitySnapshotDto.Cursor`**; the parameter is optional (`= null`) and the IB sibling omits it. Re-landing Alpaca activity paging is a product decision, not a build fix. Recorded in a comment at the site.
- **`SanitizeArtifactValue`** — restored verbatim from `716d56be`/`438673e6`, identical in both. It is genuinely distinct from the `EncodeArtifactValue` beside it: it flattens separator characters to spaces rather than quoting, which is what the call sites need.
[...]

PR #2583, opened 2026-08-08
Repair the statement connector round trip, the CI ratchets, and the accounting endpoint graph

rodoHasArrived:
## Summary

Follow-on to #2582. That PR restored the ubuntu build; this one clears the two gates that run **before** the .NET tests, repairs the defects that surfaced once tests could run, and fixes the three Windows-only compile errors that had four Windows lanes red.

**Windows desktop lane — the reason it was invisible.** `Meridian.Wpf` selects `net10.0` off Windows and puts every `ProjectReference` behind an `IsWindows` condition, so the ubuntu build compiles almost nothing from it and `Meridian.sln` never reaches the WPF sources. Forcing the Windows target framework locally reproduces exactly what the lanes hit:

```
dotnet build src/Meridian.Wpf/Meridian.Wpf.csproj -c Release \
  /p:IsWindows=true /p:EnableFullWpfBuild=true /p:EnableWindowsTargeting=true /p:WindowsPackageType=None
```

Three defects, each a merge artifact:
- `AccountingConfigureViewModel` carried a duplicated method signature line, so a declaration was followed by another declaration instead of a body (`CS1002`).
- `MainPageViewModel` resolved `LoggingService` ambiguously once both `Meridian.Wpf.Services` and `Meridian.Ui.Services` were imported (`CS0104`). Qualified the way the sibling view models in this project already do.
- `MainPageOperatingContextSelectionTests` read `Application.Current`, which binds to the `Meridian.Application` *namespace* rather than `System.Windows.Application` (`CS0234`) — the file's enclosing namespace chain reaches `Meridian`, and a namespace found that way outranks a type imported by a using directive.

Both projects now build clean at `net10.0-windows10.0.19041.0`.

**Statement connector round trip**
- The canonical artifact was writable but not readable. `ValidateAsync` reads back what the connector wrote, so the parser's accepted optional suffix now covers all ten columns the writer emits, in the same order. A narrower artifact stays readable because a shorter header is still a valid prefix.
- The writer applied two escaping conventions to one CSV record: leading columns quoted, the seven trailing columns had commas and quotes rewritten to spaces. An external transaction id used for exact matching cannot be silently rewritten, so all columns now encode.
- Statement run creation is idempotent by design — an interrupted run resumes to the same identity — so a re-import was indistinguishable from a first import and never reported as a duplicate. The connector path now checks both compatible identities before creating, matching the importer's own rule, including the canonical-only key that pre-hardening runs still carry.
- Break ids were minted with `Guid.NewGuid()`. Recovery rebuilds the match artifact when a replay resumes an interrupted run and refuses to adopt one that differs from the retained evidence, so a random id made every replay conflict with the artifact it was recovering. Ids are now derived from the break's own material.

**Reporting audit trail** — a retry is audited from the catch that decides to retry, which runs before any attempt has produced a manifest, and an audit entry cannot be appended to a run that is not retained yet. A renderer failing on its first attempt therefore never reached the terminal branch: no failed manifest was persisted and both entries were lost. Retry notes are now held until the run first exists and flushed right after the manifest is published, ahead of that attempt's own entry.

**Accounting endpoint graph** — the production-certification upsert endpoint takes its command service as a handler parameter. Minimal APIs infer a complex parameter as a service only when the container knows the type, and the endpoint test app never registered it. Parameter inference failed, endpoint-graph construction threw on the first request, and **every** route in that app answered 500 — thirty tests were failing on endpoints unrelated to certification.

**Credential catalog** — corp-actions and symbol-search identities were aliased for every provider family except AlphaVantage and Finnhub, so three credential-bearing providers resolved no descriptor at all.

**CI gates** — the file-size baseline had drifted behind six files that grew in already-merged work; `/accounting/margin-control` is now registered in the screenshot route catalog with a fixture.

## Reason

`verify-dotnet` was failing at the no-new-god-file ratchet and `verify-workflows` at the screenshot route tests, both before any test ran. Clearing them exposed the .NET failures above. The four Windows lanes had been red since 2026-08-06 with no diagnosis, because no Linux lane compiles the code that was broken.

## Testing performed

Run locally against the CI SDK pin (10.0.302):

| Check | Result |
|---|---|
| `dotnet format whitespace Meridian.sln --verify-no-changes` | exit 0 |
| `check-file-size.py` | exit 0 — 50 tracked files |
| `check-warning-suppressions.py` | exit 0 |
| `python3 -m unittest tests.scripts.test_refresh_screenshots_workflow` | 35 tests, OK |
| WPF shell at `net10.0-windows10.0.19041.0` | build 0 errors |
| WPF tests at `net10.0-windows10.0.19041.0` | build 0 errors |
| `dotnet test --filter "…Statement\|…Reconciliation"` | 3 failures, down from 15 |
| `dotnet test --filter "…AccountingSystemIntegrationServiceTests"` | 8 failures, down from 30 |

Full ubuntu lane: 81 failing tests, down from 109 before this branch.

## Known remaining failures

Stated so they are not mistaken for green. None are fixed here.

- `ScriptRunnerTests` (33) — every script run reports failure. The worker sidecar and its apphost are present in the test output, so the cause is downstream of process resolution and not yet diagnosed.
- `AccountingSystemProductionCertificationProfileEndpoint_*` (8) — these supply scope-route evidence references (`evidence://tenant/…/posting-candidate/certification`), which the Evidence Vault authority rejects as vault ids before any certification rule runs. Needs a seeded vault or a purpose-built authority seam.
- `AlpacaTradeUpdatesClientTests` (4), `ActivityFeedServiceTests` (2), `PaymentApprovalTests` (2), `OrderManagementSystemDurableHandoffOrderingTests` (2) — not yet triaged.
- `ProviderCapabilityContractRegistrationTests` (1) — `ibkr`, `nyse-streaming`, and `robinhood-live` declare a realtime capability but are never registered through `ProviderRegistry.RegisterStreamingFactory`.
- Four single-test classes — not yet triaged.

## Safety review

## Governance changes


rodoHasArrived:
### Status after `3690bb516`

**The Windows desktop lane compiles again.** Comparing the `verify-desktop` runs across the WPF fix:

| Step | `61d62ddf3` | `fdcaca0b8` |
|---|---|---|
| Restore shell / tests | ok | ok |
| Build WPF desktop shell | **failed** | ok (5m 55s) |
| Build WPF desktop tests | not reached | ok (39s) |
| Run WPF desktop development tests | not reached | **failed** |
| Run lifecycle supervisor tests | not reached | ran |

So the three compile errors were the blocker, and the lane now fails one layer deeper, on the desktop tests themselves. Those test names are only in the uploaded `windows-desktop-validation-2915` artifact, not the console, and the tests need a real Win32 dispatcher — `Application.Current.Dispatcher` — so I cannot run or read them from this Linux environment. That one is genuinely handed over rather than skipped.

**Fixed since the last push**
- `regenerate-docs` — my desktop-test edit moved the WPF screen tracker's source fingerprint, because the tracker text-scans `tests/Meridian.Wpf.Tests`. Regenerated with `scripts/generate-diagrams.mjs --tracker-only`; the value it produced (`f175940b7b9b`) is exactly the one CI computed.

**Current state**

| Lane | State |
|---|---|
| `verify-browser` | green |
| `verify-docs` | green |
| `verify-workflows` | green |
| `validate-docs`, `schema`, `scope-gate`, `Workflow Hygiene`, `Secret Scan`, `Browser W4 Parity` | green |
| `regenerate-docs` | fixed in `3690bb516`, awaiting re-run |
| `verify-dotnet` | red — the 81 documented test failures, listed in the PR body |
| `verify-desktop` + 3 Windows lanes | red — now at the WPF **test** layer, not the build |

**Not claimed as done.** The remaining `verify-dotnet` failures are enumerated in the PR description with what I know about each; `ScriptRunnerTests` (33) and the certification-endpoint cluster (8) are the two that would move the number most, and neither is diagnosed to a fix yet.

---

PR #2608, opened 2026-08-11
Adopt the 2026-08 W9 close-out delivery plan

rodoHasArrived:
## Summary

Adds a planning document and its registry records for finishing the 2026-07 first-order improvement slate. No Meridian runtime behavior changes — this is planning input only, and no roadmap row's status is advanced.

- `docs/product/w9-close-out-delivery-plan-2026-08.md` — the adopted sequence for the three open W9 rows, the ordering rationale, and a "Known constraints from source" section verified against current code.
- `docs/roadmap/data/decision-log.yml` — new `DEC-W9-CLOSEOUT-001` recording the sequencing decision.
- `docs/roadmap/data/document-index.yml` — registers the document against `W9-SAFETY-007`, `W9-GOV-008`, `W9-INGEST-009`.
- Regenerated artifacts: `docs/roadmap/generated/MANIFEST.json`, `docs/generated/repository-structure.md`, `docs/status/doc-health-dashboard.{md,json}`, `docs/status/example-validation.md`, `docs/status/coverage-report.md`.

The adopted order is `W9-SAFETY-007` → `W9-GOV-008` → `W9-INGEST-009`, delivered as eleven changes with one roadmap row advanced per change.

## Reason

The W9 slate is the registry's declared priority order and is mostly closed: `W9-ASSET-010` is `done` and six rows are `ready_for_acceptance`. Three remain open, and `DEC-DEPTH-SLATE-001` sequences the entire W10 depth slate behind them, so a partly-open slate blocks the accepted next slate. Each open row also leaves a live defect rather than a missing feature.

Checking each row against current source turned up four things worth recording before any blueprint work starts:

- **The `W9-SAFETY-007` control sweep is a WPF rewire, not a two-lane audit.** `TradingWorkspaceShellPresentationService` is the only file in `src/Meridian.Wpf/` that mentions cancel-all or kill-switch. Its Trading command bar publishes `Pause`, `Stop`, `Flatten`, `CancelAll`, and `AcknowledgeRisk`, and `ResolveActionRequest` maps every one to a pane-layout change plus a confirmation message without reaching an execution-control service. The browser lane already wires cancel-all with a disabled flag and a disabled reason, so it is the contract to mirror rather than redefine.
[...]

PR #2622, opened 2026-08-11
W9-SAFETY-007: fat-finger quantity and price-deviation rules

rodoHasArrived:
## Summary

Adds the first half of `W9-SAFETY-007`'s criterion-two rule catalogue: a fat-finger gate in the mandatory `CompositeRiskValidator`, with an absolute per-order quantity ceiling and a price-deviation band.

It follows the `OrderNotionalRule` pattern rather than inventing one — policy math in F# (`RiskRules.fatFinger`), thresholds read live from `RiskRuleRuntimeService`, and the reference price taken from `IPortfolioExposureProvider.TryGetExecutablePrice` so an order is measured at the side of the book it would actually cross.

Three design points are the substance of the change:

- **The price limb is directional.** A resting buy far below the market and a resting sell far above it are ordinary working orders, so only the aggressive side is measured — paying above the reference on a buy, hitting below it on a sell. A symmetric band would reject the desk's entire resting book, so the deviation is signed against the order's side rather than taken as an absolute. Both passive cases are asserted explicitly in tests.
- **Only the limit price is measured, never the stop price.** A protective stop sits away from the market by design, so measuring it would reject every stop-loss. A stop-limit is still measured on its limit price, because an aggressive limit is a fat finger whether or not a trigger precedes it.
- **A priced order with no reference price is refused as unmeasurable, not approved** — a band an unpriceable order sails past is not a band. It reports as unmeasurable rather than as a breach so a pricing gap cannot trip the circuit breaker, matching `OrderNotionalRule`. The quantity limb needs no price, so an unpriceable symbol does not become unroutable when only that limb is configured.

The rule runs at negative priority, ahead of the portfolio-aware ceilings, so a mistyped order is attributed to the typo rather than to whichever exposure limit its inflated size happened to breach. Severity is `Error`, not `Escalate`: a fat finger is a mistake, not a decision awaiting sign-off, so parking it would offer an operator a release for an order nobody meant to send.

Breaches carry `Code`, `ObservedValue`, and `LimitValue` rather than only a formatted sentence, so operator and audit surfaces can attribute a decision without parsing prose.

## Reason

`W9-SAFETY-007`'s second exit criterion requires the mandatory production risk validator to enforce "fat-finger quantity and price-deviation, max-notional, and price-collar rules." Max-notional already exists; this lands fat-finger. The price collar is the escalate-severity counterpart and follows in its own change — it only became expressible once the risk-engine work made severity decisional.

## Testing performed

- [x] `bash scripts/ci.sh` completed successfully — **not run in full**; the targeted suites below were run instead and the hosted `quality-gate` remains authoritative.
- [ ] Relevant integration tests were added or updated — not applicable; this adds a rule to an already-composed validator

| Check | Result |
| --- | --- |
| `FatFingerRuleTests` (new) | 16 passed |
| `Meridian.Tests.Risk` | 178 passed, 0 failed |
| risk-config / risk-endpoint / execution-governance suites | 70 passed, 0 failed |
| `dotnet format whitespace` (Meridian.Risk, Meridian.Ui.Shared) | no changes |
| `validate-roadmap-registry.py --summary` | 0 errors, 0 warnings |
| `render-roadmap-docs` / `render-source-docs` / docs-automation core | re-rendered in this commit, idempotent |
| `enforce_phase_scope.py --phase PR7` | passed |

The new tests cover both directional cases explicitly (resting buy below market and resting sell above market both approve), the stop-price exclusion, the stop-limit inclusion, both boundary conditions, the unmeasurable path, and the case where a missing reference price must **not** block because only the quantity limb is configured.

## Registry

`W9-SAFETY-007` stays `in_progress` — this discharges part of one criterion, not the row. The summary now records what actually remains: the price-collar rule, and the two-lane safety-control sweep. The sweep note is corrected to say it needs a **browser-lane inventory as well as** the WPF rewire; confirming the browser cancel-all is wired does not prove the rest of the browser safety surface is.

`validate-doc-hashes.py` reports source-hash drift for `src/Meridian.FSharp` and `src/Meridian.Ui.Shared`. That script is not wired into `scripts/ci.sh` or any workflow, and its own guidance is to refresh only after a deliberate docs/code alignment review, so it is not refreshed here.

## Safety review

## Governance changes

---
_Generated by [Claude Code](https://claude.ai/code/session_01GrqST8GCLa8rKGhxyELqPW)_


rodoHasArrived:
Round 9. The wrong-side stop finding was right and is now fixed; two P1s are deliberately not fixed here and are recorded on the roadmap row instead.

## Fixed

**Wrong-side stop triggers.** The blanket stop exclusion from the previous round was too wide. `IsStopTriggered` fires a buy once the market reaches or passes above the stop and a sell once it reaches or falls below it, so a buy stop typed beneath the market is already crossed on arrival, and a stop-market order that triggers on acceptance routes as an unbounded market order. A trigger's wrong side is the exact mirror of a limit's, so the rule now measures the mirrored direction only, under the same configured band and a distinct `FAT_FINGER_STOP_TRIGGER_WRONG_SIDE` code. A protective stop placed correctly never breaches however far away it sits — the sell stop at 40 against a 100 market still routes.

Stop-limit is no longer excluded wholesale: its trigger is measured like a stop-market trigger, and its limit against **its own trigger** rather than the market, which is what it's priced off. The sell stop at 90 with an 89 limit now reads 1.1%, not 11%. `TrailingStop` stays out — its trigger is broker-derived from `TrailPrice`/`TrailPercent` and moves with the market, so a snapshot of it measures nothing durable.

**Deviation overflow.** Confirmed at `RiskRules.fs:122` and in the C# evidence path too. Both saturate now, so `decimal.MaxValue` against a reference of 1 yields the structured breach rather than `RISK_RULE_EVALUATION_FAILED` — and the extreme *passive* sell still approves, which the saturation had to preserve.

**Metadata token mismatch.** Correct: the query asked for `fat-finger` while metadata stores `FatFinger` and `FAT_FINGER_*`. Matching now ignores separators, so a breach recorded behind a more severe rule's headline — where metadata is the only evidence — reaches its own status. This also fixed a latent case: `gross exposure` never matched `GrossExposure` either.

**Future-dated audit entries.** The window is bounded at both ends now, with a 5-minute skew allowance so ordinary host/sink drift still counts as live.

**Snapshot negatives.** Hydration rejects a negative rail rather than normalizing it to null.

**Risk module README** now documents the gate, its three limbs, and why the price limbs read `TryGetTouchPrice` rather than the valuation mark.

New coverage: 200 `Meridian.Tests.Risk` tests and a new `RiskRuleRuntimeFatFingerStatusTests` for the status projection and snapshot hydration. 2,850 tests green across the risk, UI, and execution lanes; 437 F#; `dotnet format` clean.

## Not fixed here — recorded on `W9-SAFETY-007` as named remaining work

**Fixed-income face value.** Verified: Alpaca routes bond orders with face value in `Quantity` and no notional flag to detect them the way the broker-notional exclusion does, so a 100,000-face Treasury breaches an equity-tuned ceiling while a ceiling raised for bonds leaves equities unprotected. The fix needs an asset-class seam this row doesn't otherwise touch.

**Aggressive sell amendments.** Verified: `IsRiskIncreasing` treats a limit-price *increase* as risk-increasing, but for a sell the dangerous direction is a decrease, so a sell accepted at $100 can be amended to $1 without this rule running. The fix changes shared OMS semantics for every rule, not just this one.

Both are real. Neither is a defect in the code this PR adds — they're pre-existing scope the rule newly makes visible — so they're carried in the registry rather than absorbed into a change that was scoped as "add the fat-finger rules." The row's summary now names both, with the evidence and the reason each is deferred, so they can't be silently lost or mistaken for a W9 acceptance requirement.

---

rodoHasArrived:
This PR had been invisible to CI since `4b0bfca8` (09:44), and the check results shown before now were 13 commits stale. Worth stating plainly because it changes how the earlier review rounds should be read.

The branch was conflicted against `main` (`mergeable_state: dirty`), so GitHub could not build a merge ref, and every `pull_request`-triggered workflow — `Meridian CI`/`quality-gate`, CodeQL, the WPF and route-validation lanes — simply never queued. Only `Roadmap Source Docs` ran, because it triggers on `push`. Nothing reported a failure; the runs did not exist.

Resolved by merging `main` (`01ad9aeb` → `6d221c82`). The conflicts were confined to three generated metric files — `docs/status/coverage-report.md` and both `doc-health-dashboard` outputs — where each side had recomputed the same counters. Taking either side would have been wrong, so they were regenerated on the merged tree: 2752/8781 documented items and 124,954 total lines, above both parents (2750/8780 and 2749/8780; 124,937 and 124,853), which is the confirmation that the merge carries both sets of changes.

`docs/generated/repository-structure.md` is in the merge commit for a related reason. `main` added five files without regenerating the tree doc, and `regenerate-docs` ends in `git diff --exit-code`, so the regeneration is a consequence of the merge rather than an unrelated edit. That job now passes.

No source conflicts: `main` touched none of `Meridian.Risk`, `Meridian.Execution`, `Meridian.Ui.Shared`, or `Meridian.FSharp` in that range. On the merged tree, `Meridian.Tests` builds with 0 errors and the risk/execution/UI suites are 277 passed, 0 failed; the registry validates with 0 errors and the PR7 scope gate passes.

The full suite is running now for the first time since this morning, so treat this as the PR's first real CI signal.

---

rodoHasArrived:
`quality-gate` is green on `f8ae01d8` — `verify-dotnet`, `verify-browser`, `verify-docs`, `verify-workflows` all pass, along with `Analyze csharp`, `Pilot Acceptance Evidence`, both route validations, and the WPF lanes. That is this PR's first authoritative pass, for the reason in my previous comment.

CodeQL is the one red check, and I'm not changing code for it. The reasoning, since four high alerts on a risk-control PR shouldn't be waved past:

**The alerts.** Four instances of `cs/cleartext-storage-of-sensitive-information` at `FatFingerRule.cs` lines 191, 238, 285, 306. All four are the same argument — `LogSanitizer.Sanitize(request.Symbol)` — and the reported source is *"data returned by call to method `RemoveBrokerAccountAndOverrideKeys : OrderRequest`."*

**Why it is backwards.** That method is the redaction. `ExecutionOrderMetadataPolicy.RemoveBrokerAccountAndOverrideKeys` strips `alpaca:broker_account_id`, `manualOverrideId`, `executionControlOverrideId` and their snake_case aliases from the request metadata, and `OrderManagementSystem.cs:371` passes its output — `safeRequest`, not `request` — into `_riskValidator.ValidateOrderAsync`. So the object CodeQL is calling sensitive is by construction the one with the sensitive keys already removed, and the query has latched onto the method's name rather than its effect. The only value these four calls log is `Symbol`: a ticker. The rule logs no metadata, no account id, no override id.

**`LogSanitizer` did what it was for.** It is a barrier for the log-injection query (`Sanitize` neutralizes control characters via `String.Replace`, which is the form static analysis recognizes), and the log-entries-from-user-input findings that motivated adding it are gone. It is not a barrier for the cleartext-storage query, which models sensitivity by source rather than by escaping — a different query with a different sanitizer model. Adding it was still correct; it just never addressed this one.

**It is the established pattern, not a new one.** `LogSanitizer.Sanitize(request.Symbol)` appears 14 times across 8 files, 10 of them already on `main` — including `PositionLimitRule.cs:82` and `DrawdownCircuitBreaker.cs:61`, which are rules in this same `CompositeRiskValidator` chain receiving the same `safeRequest`, and `OrderManagementSystem.cs:215`, which logs the symbol immediately after calling the very method CodeQL names as the source. Only my four are reported because the check scopes to code changed by the pull request.

I'd rather leave four explained alerts than contort the code around a naming heuristic. The alternatives are worse: dropping the symbol from these logs removes the one field that makes a rejection diagnosable, and renaming a shared `Meridian.Execution.Sdk` method to dodge the heuristic would touch the SDK and its callers for no behavioural reason. **I have not dismissed the alerts** — that is a security-governance call for a human, and the evidence above is what it should be made on.

Separately, the check's stated title is `2 configurations not found` (`/language:actions`, `/language:python` present on `main` but not in this run). That is a repo-level scanning-config gap, not a finding against this diff.

---

rodoHasArrived:
Two review findings on `fc834414`. One is incorrect and I'm not changing code for it; the other is real but belongs to change 3, and is now recorded on the row.

**`Math.Abs(decimal.MinValue)` does not throw.** The finding says the quantity magnitude at `FatFingerRule.cs:324` raises `OverflowException` before the structured check runs, so a `decimal.MinValue` quantity would surface as `RISK_RULE_EVALUATION_FAILED` instead of `FAT_FINGER_QUANTITY_EXCEEDED`. That reasoning holds for the integral types, whose ranges are asymmetric — `Math.Abs(int.MinValue)` genuinely throws. `decimal` is not one of them. Run on the SDK in this container:

```
decimal.MinValue = -79228162514264337593543950335
decimal.MaxValue =  79228162514264337593543950335
symmetric: True
Math.Abs(decimal.MinValue) = 79228162514264337593543950335  -> NO THROW
```

`decimal.MinValue` is exactly `-decimal.MaxValue`, so the magnitude is representable and no saturation is needed. The overflow that *is* reachable — `quantity * largestRatio`, where the gateway bounds ratios below but not above — was already saturated at lines 342–344, which is the behaviour the finding correctly points at as the model.

**The browser cannot enable either threshold, and that one is real.** `risk-control-panel.view-model.ts:84` binds a single hard-coded `DRAWDOWN_RULE_NAME = "DrawdownCircuitBreaker"` and renders only the drawdown editor, so an operator sees the fat-finger status row with no control that turns it on — and both thresholds default to null, meaning the shipped rule approves everything until configured.

I'm not adding those controls here, for a reason worth stating rather than asserting: `maxOrderNotional` — the sibling threshold on the same `RiskRuleConfig`, belonging to `OrderNotionalRule`, the rule this one is patterned on — is *also* absent from the panel and appears only in the TypeScript contract. So the panel has never edited a threshold of this kind; my two fields land with exactly the posture the existing one has. This is a pre-existing shape in the panel, not something this change introduces, and it is precisely what change 3's two-lane safety-control sweep exists to audit — every surfaced control either wired to a real service or explicitly demoted.

It is now named in `W9-SAFETY-007.current_summary` so the criterion cannot be discharged over it.

**One correction to the row while I was there.** The same commit retracts a justification this row asserted: that `LimitOnOpen`/`LimitOnClose` contribute no price because they price against a future auction. True about the reference, false about the risk — `MapOrderType` maps both to a plain broker-enforced `limit`, `MapTimeInForce` carries the auction timing separately as `opg`/`cls`, `BuildOrderPayload` writes `limit_price` unconditionally, and `BrokerageGatewayAdapter` *requires* a positive limit price for both. A $1,000 buy limit on a $100 symbol routes today. It is recorded as a gap rather than a correct exclusion.

---

rodoHasArrived:
Five findings on `0699c2b5`. One is factually wrong and now has a test proving it; three are gaps already recorded on the row with an owner; one restates a bound the row already carries. Taking them individually so none of it rests on "already handled."

**`decimal.MinValue` does not throw — now pinned rather than asserted** (`FatFingerRule.cs:330`). This was raised earlier against `request.Quantity` and is now raised against `leg.RatioQuantity`. The premise is the same and it is the same error: the `Math.Abs(MinValue)` hazard is real for the *integral* types, whose ranges are asymmetric, but `decimal.MinValue` is exactly `-decimal.MaxValue`, so `Math.Abs` is total over `decimal` and cannot overflow. The saturation already present then keeps the product finite — a `MaxValue` ratio drives the guard to `quantity > 1`, and anything past it returns `decimal.MaxValue` instead of multiplying.

Since answering the same claim twice in prose is not evidence, `47357d90` pins it as a theory across **both** arms of `ResolveEffectiveQuantity`, which feed different paths:

```
Quantity_AtTheDecimalExtremes_ReportsTheStructuredBreachRatherThanThrowing(onTheLeg: True)
Quantity_AtTheDecimalExtremes_ReportsTheStructuredBreachRatherThanThrowing(onTheLeg: False)
```

Each asserts no `OverflowException`, then the structured `FAT_FINGER_QUANTITY_EXCEEDED` with `ObservedValue` saturated to `decimal.MaxValue` against the configured limit — exactly the outcome the finding asks for, already produced. 69 fat-finger tests pass, 281 across the affected suites. Worth a test rather than a reply because the input *is* reachable: the OMS validates risk before the gateway validates the request, so the positive-whole-ratio contract has not been enforced when this rule runs.

**Auction limits** (`:355`) — agreed, and already recorded as a defect rather than a correct exclusion. I verified the same mechanism independently: `MapOrderType` maps both types to a broker-enforced `limit`, `MapTimeInForce` carries the auction timing separately as `opg`/`cls`, `BuildOrderPayload` writes `limit_price` unconditionally, and `BrokerageGatewayAdapter` *requires* a positive limit price for them. `W9-SAFETY-007.current_summary` now states this is a gap, and the delivery plan — merged to `main` in #2608 — assigns change 2 the auction-appropriate reference with a fail-closed unmeasurable fallback. Implementing it here would duplicate that change's scope; the criterion is explicitly not dischargeable while it stands.

**Fixed-income face value** (`:179`) and **amendment revalidation** (`:151`) are likewise recorded, not deferred silently. The amendment one is the more serious and the row says so: `IsRiskIncreasing` treats only a numerically *higher* limit or stop as risk-increasing, while the dangerous direction is a *decrease* for a sell limit and for a buy stop alike — a sell accepted at $100 amended to $1, or a buy stop accepted at $110 amended to $1 and triggering immediately as an unbounded market order. That is a bound on what these limbs guarantee, and the row states they hold on **submission only**. It is change 2's because the fix changes shared OMS semantics for every rule, not just this one, and that is a decision to take deliberately rather than as a side effect of adding a rule.

Distinguishing the two dispositions, since it is the whole point: the `decimal` finding is refused because it is untrue, and now has executable proof. The other four are accepted, written into the registry row so a registry-first implementer meets them, and assigned to a change that cannot mark the criterion met while they stand — which is not the same as deferring them.

---

rodoHasArrived:
Two of the three findings on `19a61850` are fixed in `44e592e5`. The third I'm declining, and the reason matters more than the decision.

**Uncapped amendments (P1) — correct, and it makes the rule simpler.** My previous condition required the measured comparison *and* a non-increasing quantity, so an uncapped order repriced upward at the same size still took the full-order branch. For an uncapped order the entered price is not what it pays, so measured value is not exposure **at all** — only quantity is, because the mark is common to both sides of the amendment. The two cases now use the evidence each actually supports:

- capped buy — the limit is the price paid, so the increment is the measured difference floored at zero;
- everything else — zero unless quantity rose; if it rose, the increase is Δquantity × mark and the OMS has no mark, so the whole order goes to the rules.

Third round on this function, and the shape of my error each time was adding a qualifier to a test that was measuring the wrong thing, rather than asking what the measure means for that order.

**Future records holding the cap (P2) — correct.** Bounding their effect on gap *reporting* left them able to fill the cap, and since they sort after every real entry a "keep the newest" trim drops them last, so once full, every genuine entry arriving afterwards became the oldest and was discarded on arrival. The trail stopped retaining current activity while each discard refreshed the incompleteness timestamp. They are now evicted before the cap applies, kept only while there is room to spare so ordinary clock skew loses nothing.

That change also obsoleted one of my own tests: with future records evicted first, its 1,500 real entries now fit and it no longer produced the gap it asserted. It now uses 2,500 real entries to overrun the ceiling on their own, which preserves what it was for — eviction must not *hide* a real gap.

**The notional-metadata gate ordering (P2) — declining, and I reverted a change I had already made.** The finding is right that `OrderNotionalResolver` ignores notional metadata under the face-value marker (`OrderNotionalResolver.cs:203-207` says so explicitly), so measurement would agree and nothing would diverge. I reordered the gate on that basis, and it failed `PlaceOrderAsync_WhenTheGatewayRoutesQuantityForThisAssetClass_RejectsNotionalMetadata` — which asserts the refusal deliberately, with its reasoning in the test body.

On reflection that test is right and my change was wrong. The argument for accepting covers measurement; it does not cover intent. A caller who wrote `notional=1` on a `Quantity=5000` treasury order asked for a **one-dollar order**. Accepting it so the gateway routes 5,000 face value honours something they did not request, and does so silently. Refusing tells them to correct the request, which is what "refusing beats measuring one size and routing another" exists for. Reordering would need the caller's intent addressed too — stripping the metadata explicitly, or rejecting the contradiction rather than the shape — not just proof that the rails happen to agree.

670 tests pass across risk, execution, audit, OMS, paper and Alpaca; format clean; PR7 scope gate passes.

---

rodoHasArrived:
Both findings on `220844dc` are correct. One is fixed in `d7f79532`; the other I've deliberately not fixed, and it needs a decision before this PR merges.

**Legs snapshot — fixed.** Metadata was not the only caller-owned collection on the request. `Legs` is a list the caller may still hold, and the fat-finger quantity limb multiplies by the largest leg ratio, so a ratio raised while validation is awaiting routes more contracts than the ceiling approved. Snapshotting one collection and not the other only moved the race. Both are copied at the same boundary now, with a test that mutates the caller's list from inside the validator.

That is the third time in this PR I have fixed an instance and left its category, so it is worth naming rather than quietly patching: the snapshot commit reasoned about `Metadata` because `Metadata` was what the finding named, not about "caller-owned mutable state on the request", which is what the property actually is.

**Percentage-of-par vs fill economics — not fixed, and recorded as blocking.** Verified as described:

- `TradeExecutedEvent.GrossValue` is `FilledQuantity * FillPrice`, unscaled
- `LedgerPostingConsumer` posts that value straight to the securities and cash accounts
- `PaperTradingPortfolio.ApplyFill` takes a `contractMultiplier` but no face-value flag

So 100,000 face at 101.25 is admitted and measured at **$101,250** by the three scaled sites, then books **$10,125,000** into cash, positions, exposure and the ledger. A hundred times the economic value, on an order pre-trade validation explicitly endorsed at the correct size.

I did not fix it here for a scope reason I want to be explicit about rather than have read as deferral. Carrying the sizing classification through the execution report into fill processing and accounting economics is the face-value feature's own work — it touches portfolio and ledger posting, not the rule catalogue this row's criterion is about — and it needs round-trip proof that a face-value fill books what pre-trade measured. Rushing a ledger-posting change at the tail of a long session is how a 100× error gets a second one layered on it.

What `W9-SAFETY-007` now records is deliberately stronger than "remaining work": **the asymmetry is worse than scaling neither side**, because the discrepancy carries a passed risk check in front of it, so the face-value sizing path is unshippable rather than incomplete. If that path is meant to ship with this PR, that gap has to close first; if it is not, it should be gated off rather than left half-scaled.

672 tests pass; file-size ratchet passes; registry validates; PR7 scope gate passes.

---

rodoHasArrived:
Correct, and already recorded and gated — no change on `d7f79532`, so this one is a pointer rather than a fix.

`W9-SAFETY-007.current_summary` carries it verbatim, including the market-primary case this finding sharpens:

> Attached bracket, OCO, and OTO legs are also unmeasured — the rule reads the top-level `LimitPrice` and `StopPrice` only, while `AlpacaBrokerageGateway` routes operator-entered `take_profit.limit_price`, `stop_loss.stop_price`, and `stop_loss.limit_price` from metadata, so **a market bracket order bypasses the price limbs entirely** and a sound parent can carry a severely mistyped exit that activates the moment it fills; the fix reads the same metadata aliases the gateway reads and is deferred because it is a broker-specific seam on order classes this row's limbs were not scoped to.

And the delivery plan — merged to `main` in #2608, so it is authoritative rather than a proposal — assigns it at lines 267–273 under the heading *"The same holds for broker-normalized child legs, **and it changes what change 2 may discharge**"*, requiring change 2 to resolve and validate every gateway-normalized child price through the same metadata aliases the gateway reads *before* it marks the pre-trade rule criterion met.

That is the distinction I have been trying to hold throughout this PR: the criterion is gated over this, not merely annotated with it. Change 2 cannot record the rule catalogue as complete while attached child prices are unmeasured.

Worth stating where this leaves the PR overall, since this is the tenth review round: every finding still outstanding against it is now either **gated to change 2** (auction limits, fixed-income face value, amendment approval path, bracket/OCO child legs), **a decision for a human** (the CodeQL cleartext family, which my last comment explained and which grew from 4 to 34 only because my file-size decomposition moved pre-existing log calls into the diff), or **the one blocker I raised myself** — percentage-of-par reaching pre-trade measurement but not fill economics, which books a fill at 100× and should be resolved or gated off before this merges.

---

rodoHasArrived:

rodoHasArrived:
Conflicts are resolved in `01a88927` (merged `main` at `89d29af3`). I went ahead rather than leaving it, because this particular conflict has a resolution that looks wrong if you treat it as an ordinary content conflict — flagging that here in case Copilot or anyone else takes a pass at it.

**All three conflicts are generated status artifacts** — `coverage-report.md` and both `doc-health-dashboard` outputs — where each side recomputed the same repo-wide counters. **Picking a side is wrong for all of them.** Neither branch's numbers describe the merged tree, and `regenerate-docs` ends in `git diff --exit-code`, so a side-picked resolution fails CI even though the merge itself looks clean.

The correct resolution is to clear the markers and then regenerate:

```bash
git checkout --theirs docs/status/coverage-report.md docs/status/doc-health-dashboard.json docs/status/doc-health-dashboard.md
```

The regenerated total (126,748 lines) lands above both parents, which is the check that the merge actually carries both sides.

This is the fifth merge on this branch and the fifth time these same three files have conflicted. It is structural rather than incidental: the dashboards embed repo-wide and per-file line counts, so **any** merge from `main` invalidates them regardless of what either side changed. Worth fixing at the generator or gate level — either by not tracking volatile counters in version-controlled artifacts, or by regenerating them in CI rather than asserting they were committed — but that needs `build/scripts/**`, which is outside this PR's phase scope.

Verified on the merged result: registry validates, file-size ratchet passes, PR7 scope gate passes, build clean, 672 tests pass across risk, execution, audit, OMS, paper and Alpaca. `main` did touch one file in my areas this time (`73247d6e`, shared text primitives in `LedgerPostingConsumer`) — no behavioural overlap, and the suites confirm it.

---

rodoHasArrived:

Issue #2620, opened 2026-08-11
Codebase audit (Aug 2026): findings index and prioritization


rodoHasArrived:
Tracking issue for a codebase-wide audit covering duplication, understandability, modularity, robustness, extensibility, and pragmatism.

**Scope audited:** 4,354 C# files / 260k LOC in `src/`, ~299k LOC TypeScript in the browser workstation, plus F#, Python, and XAML surfaces.

> **Revision note.** #2610 and #2611 were both revised after a second verification pass. #2611's original correctness claim was **wrong and has been retracted** (no drift bug; the duplication finding stands). #2610's severity was **softened** from "demonstrated live defect" to a confirmed-ingredients latent trap — but the verification pass also turned up materially stronger evidence, so the issue is now more actionable than it started. The table below reflects the corrected state.

## Findings

| # | Finding | Dimension | Severity |
|---|---|---|---|
| #2610 | SHA-256 digest casing inconsistent — uppercase producers exist, some validators reject uppercase | Robustness / duplication | **High — confirmed producer bug, latent blast radius** |
| #2612 | InMemory/Postgres service pairs duplicate business rules, including data-scoping predicates | Duplication / robustness | **High — drift risk** |
| #2614 | No shared string primitives — `NormalizeOptional` in 62 files, `RequireText` in 31 | Duplication / pragmatism | Medium — systemic root cause |
| #2611 | Ledger dimension helpers duplicated across 7 files, incl. a copy in the endpoint layer | Duplication / modularity | Medium — no live defect |
| #2613 | `EnsureHumanOrigin` governance gate copy-pasted across 15 services with 3 exception types | Duplication / extensibility | Medium |
| #2615 | No clock abstraction — 2,450 `DateTime.UtcNow` vs 167 `TimeProvider` | Robustness / testability | Medium |
| #2616 | Provider integration transform helpers duplicated across three mapping services | Duplication / extensibility | Medium |
| #2617 | Dashboard TypeScript `strict` off; only `strictNullChecks` gated, tests excluded | Robustness / extensibility | Medium |
| #2619 | God-file ratchet freezes 169k LOC across 50 files with no burn-down plan | Modularity | Medium — systemic |
| #2618 | Endpoint error handling inconsistent — `EndpointHelpers` in 46 of 143 files | Pragmatism / consistency | Low-Medium |

## Suggested sequencing

[...]

PR #2674, opened 2026-08-11
Make the god-file ratchet report its trend, and set burn-down targets

### LINKED ISSUE #2619: God-file ratchet freezes 169k LOC across 50 files but has no burn-down plan
Found during a codebase-wide audit (modularity).

## Summary

`build/scripts/ci/check-file-size.py` is a good mechanism — it prevents *new* god files and stops existing ones growing. But it is a containment tool, not a reduction tool. The baseline currently freezes **169,329 lines across 50 files**, and nothing drives that number down. This issue proposes decomposition targets, not a change to the ratchet.

## Evidence

`build/config/file-size-baseline.json` — threshold 2,000 lines, 50 baselined files, 169,329 LOC total.

Worst offenders:

| LOC | File |
|---:|---|
| 7397 | `src/Meridian.Ui/dashboard/src/screens/settings-screen.tsx` |
| 7147 | `src/Meridian.Ui/dashboard/src/screens/accounting-screen.view-model.ts` |
| 6673 | `src/Meridian.Ui/dashboard/src/lib/dev-fixtures.ts` |
| 6126 | `src/Meridian.Ui/dashboard/src/screens/accounting-screen.tsx` |
| 5358 | `src/Meridian.Wpf/ViewModels/Accounting/AccountingConfigureViewModel.cs` |
| 5243 | `src/Meridian.Ui/dashboard/src/screens/trading-screen.view-model.ts` |
| 4740 | `src/Meridian.Ui.Shared/Services/SecurityMasterWorkbenchQueryService.cs` |
| 4478 | `src/Meridian.Ui.Shared/Endpoints/WorkstationEndpoints.cs` |
| 4408 | `src/Meridian.Wpf/ViewModels/SecurityMasterViewModel.cs` |

Structural measures on the C# side make the coupling concrete:

| File | LOC | public members | injected deps |
|---|---:|---:|---:|
| `AccountingConfigureViewModel.cs` | 5357 | **224** | **18** |
| `SecurityMasterViewModel.cs` | 4408 | **212** | **20** |
| `FundLedgerViewModel.cs` | 3325 | 152 | 14 |
| `MainPageViewModel.cs` | 2313 | 134 | 19 |

`AccountingConfigureViewModel` takes 18 constructor dependencies, 13 of which are nullable (`ICapitalAccountWorkbenchService?`, `IAccountingPolicyService?`, `AccountingProductionReadinessService?`, …) and exposes 48 `ObservableCollection` properties. It has no `#region` structure — 5,357 lines with no internal seams.

On the frontend, `settings-screen.tsx` holds 18 component/function declarations, 22 `useState` calls, and 11 `useEffect` calls in one file.

## Why it matters

**The ratchet's own rationale is the argument.** Its header cites ADR-017 (modular operational monolith) and ADR-018: *"capability logic belongs in composed, single-responsibility units, not god files."* The baseline records where that is currently untrue; without a burn-down it just makes the debt permanent and audited.

**Nullable-dependency god classes hide their real contract.** 13 optional dependencies means ~2^13 possible service configurations, and every method has to defensively re-check which collaborators exist. Nothing in the type system says which combinations are valid, so the constructor stops describing what the class needs.

**224 public members on a ViewModel is not a testable unit.** These files are the ones a change is most likely to touch and least likely to be safely reviewed — 5,000-line diffs are read, not reasoned about.

**The ratchet has a cliff.** Any file at its recorded cap fails CI on a one-line addition, forcing either an unrelated refactor or a `--update-baseline` commit. `AccountingConfigureViewModel.cs` is currently at 5,357 against a 5,358 cap — one line of headroom. That pressure tends to produce baseline bumps rather than decomposition. (Issue #2276 already reports the god-file ratchet lane red on `main`.)

## Suggested direction

1. **Set a burn-down target** — e.g. reduce baselined total by 15% per quarter, or retire N files from the baseline per release. Without a number this stays static.
2. **Start with the highest-coupling C# ViewModels**, not the largest files. `AccountingConfigureViewModel` (18 deps / 224 members) and `SecurityMasterViewModel` (20 deps / 212 members) are the best value: split by workflow area into child ViewModels, each with its own non-nullable dependency set.
3. **`settings-screen.tsx` and `accounting-screen.tsx`** decompose naturally along the tab/section boundaries already present in the UI.
4. **`dev-fixtures.ts` (6,673 LOC)** is a different case — it is dev-only, dynamically imported behind `import.meta.env.DEV`, and correctly tree-shaken from production. Consider excluding fixture files from the ratchet rather than spending refactor effort there.
5. **Extend the ratchet to report the trend**, so the baseline total appears in CI output and regressions are visible as a number rather than a per-file pass/fail.

rodoHasArrived:
## Summary

Adds `docs/development/god-file-burn-down-plan.md` with measured targets and decomposition sequencing, and makes the ratchet report the number that plan is trying to move.

**Reporting only — no enforcement rule changes**, so nothing that passes today starts failing. The committed baseline is untouched.

| Change | Effect |
|---|---|
| Trend line on every run | Tracked files, capped lines, current lines, reclaimable lines |
| `TIGHT` warnings | Files within 25 lines of their cap, worst first, distinguishing the ones actually pinned |

> **Scope note:** this PR originally also added a `--tighten-baseline` flag. It was **withdrawn** after review found six distinct defects in it across five rounds, with none anywhere else in the change. The mechanism and its design constraints are now tracked in **#2675**. See "Why the flag came out" below.

## Reason

Fixes #2619.

### The measurement changed the plan

The issue notes `AccountingConfigureViewModel.cs` at 5,357 against a 5,358 cap and calls it a cliff. Measuring the whole baseline, it isn't a cliff — it's the steady state:

```
Baseline trend: 50 tracked file(s), 169,329 capped line(s),
                169,328 current line(s), 1 line(s) reclaimable.
TIGHT: 50 of 50 tracked file(s) are within 25 line(s) of their cap. 49 of them sit at
their cap, where a single added line fails this check. Decompose before extending.
```

**49 of 50 files have exactly zero headroom.** Total reclaimable slack across 169,329 tracked lines is **one line**.

So adding a single line to almost any god file fails CI right now. That systematically pushes ordinary maintenance toward `--update-baseline`, and the ratchet ends up recording growth rather than preventing it. Not hypothetical — it happened three times during this batch of work. PR #2669 *removed* 91 lines from a sibling partial class and still failed, because one added `using` took `LedgerEndpoints.cs` from 2,824 to 2,825 against a 2,824 cap.

### Two corrections to the issue's framing

- **The browser workstation is 48% of the baselined debt** (21 files, 81,470 lines). A C#-ViewModel-only effort leaves half the problem standing.
- **The `dev-fixtures.ts` exclusion argument does not hold as stated.** Both #2619 and the audit describe it as dev-only and dynamically imported behind `import.meta.env.DEV`. But `src/Meridian.Ui/dashboard/vite.config.ts:6` imports `resolveDevFixture` **statically**, and that config is evaluated by `vite build`; the preview proxy calls it at line 80. The fixtures may still be absent from the emitted bundle, but the dev-only premise was the load-bearing half of the argument for excluding 6,673 lines from a guardrail. The plan now says so and recommends deciding on effort grounds instead. **No exclusion is made here.**

### Targets are proposed, not committed

`docs/roadmap/data/` is the authoritative planning registry and holds no file-size item. Stating per-release and per-quarter numbers only in a supporting plan would create roadmap truth outside the registry, so the table is explicitly marked proposed, with a registry entry named as the adoption step. I have not added that entry — minting a tracked commitment with a wave, lane and priority is a planning decision, not a cleanup side effect.

### Why the flag came out

Five review rounds found six defects in `--tighten-baseline`/`--buffer`: retired files dropped from the reclaimed total; retirement discarded a cap while the file was still inside the buffer, leaving it to fail as a *new* god file; an unreadable tracked file counted as empty and lost its cap; an unreadable *untracked* file was skipped so tightening could miss a new god file; the command returned 0 while the ratchet was failing; and retained buffer read as reclaimable slack, so the tool recommended destroying its own headroom.

All six sit in one seam — the interaction between *file below threshold*, *file still in baseline*, and *tree currently failing*. A read-only check never reasons about it; a mutating one hits it on every path. Fixing the last one properly likely needs a baseline schema change to distinguish deliberate headroom from unlocked reduction, which is a format decision rather than a patch.

Keeping it here would have held up the reporting indefinitely. #2675 carries the requirement list and the numbered defects as acceptance criteria.

## Testing performed

> Unusually for this batch, this change is **fully verified locally** — Python and docs only, no .NET involved.
>
> - script-test lane via `build/scripts/ci/run-script-tests.py`: **766 tests, 0 failures**
> - `python3 build/scripts/ci/check-file-size.py` — OK against the real baseline, output above
> - docs rules engine PASS, AI inventory pass, `run-docs-automation --profile core` re-run
> - scope gate passes with **no phase marker** — nothing here is on the governed authoring surface

Tests in `tests/scripts/test_check_file_size_ratchet.py` — the first this script has had — cover trend totals, a baselined file that shrank below the threshold (counted from disk rather than as zero), `TIGHT` firing and not firing, the pinned-vs-approaching distinction, trend reporting on failure, and that a brand-new oversized file still fails.

They live in `tests/scripts` because that is `run-script-tests.py`'s discovery root; `build/scripts/tests`, where they started, is discovered by no CI lane.

## Safety review

## Governance changes

`build/scripts/ci/check-file-size.py` is invoked by `scripts/ci.sh` but is not on the template's governance list, and no enforcement behavior changes — the added output is informational.


rodoHasArrived:
All five review points were correct. Fixed.

### Count below-threshold files in the trend totals

The sharpest catch — the totals were wrong for exactly the case the plan cares about. A baselined file that shrinks below the threshold drops out of `_scan`, so it was counted as 0 current lines and 0 reclaimable. Your example is precisely right: capped at 2,010 and reduced to 1,999 reported as 0/0 rather than 1,999/11. The metric broke at the moment a decomposition was about to succeed.

Line counts now come from disk for any baselined file the scan omits. Test added covering a file that shrank below threshold.

### Reject threshold changes when tightening

Real bug in the new flag. `--threshold 40 --tighten-baseline` scanned at 40, omitted still-protected files, and the retirement filter deleted their caps — a "downward-only" operation silently dropping protections, exiting 0. It now refuses a threshold other than the baseline's recorded one, with a test.

### Do not describe tightening as restoring headroom

You're right, and it went deeper than wording: the plan's *immediate target* was self-contradictory. Tightening to the exact current count re-pins the file, which is the state the plan exists to escape, so "restore working headroom via `--tighten-baseline`" could never be true.

Rather than only reword it, I added `--buffer N`: `--tighten-baseline --buffer 25` locks in the reduction while leaving 25 lines of working room, and still never raises a cap above its existing value (`min(cap, current + buffer)`). The immediate target now reads as that command. Three tests, including one that the buffer cannot be used to raise a cap on a file that hasn't shrunk enough.

### Do not warn that one added line always fails

Correct, and the current baseline made it immediately false — one file has a spare line. The warning now reports how many files sit *at* their cap and reserves the one-line-fails wording for those:

```
TIGHT: 50 of 50 tracked file(s) are within 25 line(s) of their cap. 49 of them sit at
their cap, where a single added line fails this check. Decompose before extending.
```

### Put the ratchet suite under CI discovery

Right, and better than what I wrote in the PR body — I'd assumed wiring these in meant editing `.github/workflows/**` and left them unrun. `run-script-tests.py` discovering `tests/scripts` is the seam I missed. Moved there; the script lane now reports **775 tests, 0 failures** (16 mine, up from 13 with the new cases).

### ADR-018 citation

Correct — ADR-018 is `018-declarative-statement-mapping-profiles.md`. I copied the misattribution from the ratchet's own module docstring, which says "ADR-018 (module conventions)". The plan now links [ADR-017](../adr/017-modular-operational-monolith.md) and [module conventions](../architecture/module-conventions.md) directly.

Worth noting the script's docstring carries the same wrong citation on `main`. I've left it, since correcting it is unrelated to this change and belongs in its own commit — happy to fold it in if you'd rather.

---

rodoHasArrived:
Third round — all three correct again, and all three were bugs in the tightening path this PR introduced. Fixed.

### Include retired files in the reclaimed-line total

The sum only walked surviving entries, so retiring a file discarded its reduction. Your example exactly: capped at 30, cut to 5 → `25 line(s) reclaimable` before, `0 line(s) reclaimed` after. The metric went blank precisely when a decomposition achieved the plan's *headline* goal — retiring a file — which is about the worst place for it to under-report. Retired entries now contribute their cap-to-current reduction.

### Preserve the requested buffer before retiring a file

The subtlest of the three, and the one with the worst failure mode. Retiring removes the cap, leaving only the threshold as protection — so a file capped at 2,100 and reduced to 1,999 was retired by `--buffer 25` and sat one line under 2,000. The next added line wouldn't hit a cap, it would fail as a **brand-new god file**: a harder failure than the cap it just lost, and the exact opposite of what the buffer promised.

Entries are now held at `min(old_cap, current + buffer)` until the threshold itself supplies the requested room. Two tests: one that a file at 9 lines with `--threshold 10 --buffer 5` keeps a cap of 14 and is *not* retired, one that a file at 3 lines does retire because the threshold already covers the buffer.

### Report only the headroom actually retained

Right — the message reported the request, not the outcome. A file at 29 under a cap of 30 printed `25 line(s) of headroom kept` while having one. It now reports the smallest amount actually retained and names the request separately:

```
Tightened baseline: 1 line(s) reclaimed, 0 file(s) retired, 1 still tracked,
as little as 1 line(s) of headroom retained (requested 25; existing caps limited the rest).
```

### Also

Factored the live-line lookup out of `_report_trend` into `_live_lines`, so the check path and the tighten path cannot drift on how they count a file the scan omits — that shared assumption was the root of both this round's accounting bug and last round's trend bug. And softened the plan's `--buffer` description, which promised 25 lines of room unconditionally; it now states that an existing cap can limit it and that entries are held rather than retired until the threshold covers the buffer.

Script-test lane: **779 tests, 0 failures** (20 in this suite, up from 16).

Worth saying: three rounds on this PR have each found real defects in the tooling change, and the pattern is consistent — every one lived in the interaction between "file below threshold" and "file in baseline". That is the seam the original script never had to think about, because it only ever asked one question per file.

---

rodoHasArrived:
Fourth round — four findings, all correct, all fixed. Two of these were the most serious yet.

### Fail tightening when violations remain

Tightening wrote the baseline and returned 0 *before* the violation checks ran. So running it in a tree with a new god file or a grown file exited successfully while the ratchet was failing — and the write baked the violation in as the new normal. Worse in context: the plan tells people to run this command after every decomposition, so the false green sits directly on the recommended path.

Violations are now computed before the branch, and tightening refuses with exit 2 and lists them. Two tests: a grown file, and a new god file with the baseline verified unchanged.

### Abort tightening when a tracked file cannot be read

`_count_lines` swallowing `OSError` is fine for reporting and wrong for a mutating path: a file that exists but can't be read looked empty, so tightening retired it and wrote its cap away. Silent loss of a protection, triggered by a transient I/O error.

`_live_lines` now takes a `strict` mode used only by tightening, which lets the error surface and aborts with exit 2. A genuinely absent file still retires normally — that distinction is the one you asked for. The test chmods a tracked file to `0o000`; when the runner is root and permissions don't bite, it **skips** rather than asserting a false pass.

### Reject buffer outside tightening mode

`--update-baseline --buffer 25` exited 0 while pinning every cap to the current count — exactly opposite to the request, and committable. Now rejected with a message naming that specific trap.

### Qualify the one-line failure claim

Right, and I'd half-fixed it: I corrected the `TIGHT` output last round but left the same unconditional claim in the plan prose. The ratchet rejects only `lines > cap`, so the one file with a line spare reaches its cap and passes. Scoped to 49 of 50.

### One test changed meaning

`test_never_raises_a_cap_when_the_file_grew` asserted exit 0 on a grown file, which is now the refusal path. Rewritten as `test_refuses_to_tighten_while_a_file_has_grown`. Flagging it explicitly because a test whose expectation flips deserves a look rather than a silent edit — the underlying property (never raise a cap) is still covered by the `--buffer 50` case on a 29/30 file.

Script-test lane: **782 tests, 0 failures** (23 in this suite, 1 skipped).

---

Four rounds, fifteen findings, and I'd note the shape: every single one has been in code this PR added, and the tighten path has now been wrong in four distinct ways — accounting, retirement, error handling, and exit contract. That is a fair signal that a mutating flag on a CI guardrail was more than it looked. If you'd prefer to land the reporting and the plan on their own and take `--tighten-baseline` as a separate change with its own review, I think that's a defensible call and I'm happy to split it.

---

rodoHasArrived:
Fifth round. Two of the four findings are again in `--tighten-baseline`, which settles the question I raised last round: **I've withdrawn the flag from this PR.**

### Why, concretely

Five rounds, six defects, all in the mutating flag; zero code defects anywhere else. The two newest continue the exact pattern:

- **Unreadable untracked source** — I hardened the strict read for *baseline* entries last round, but `_scan` still counts an unreadable untracked file as zero, so tightening could rewrite the baseline while missing a new god file entirely. Same bug, one layer out.
- **Buffer vs. slack** — headroom a buffer deliberately retains is counted as reclaimable, so the next ordinary run recommends the command that destroys it. The tool undermining its own recommended workflow.

Every one of the six lives in the same seam: the interaction between *file below threshold*, *file still in baseline*, and *tree currently failing*. A read-only check never has to reason about that. A mutating one hits it on every path. That's design work, and it was never going to converge inside a reporting PR.

### What's left

What #2619 actually asked for, and what has drawn no findings in five rounds:

- trend totals (tracked files, capped lines, current lines, reclaimable)
- `TIGHT` warnings with the pinned-file distinction
- the burn-down plan

The plan now has a **"The missing mechanism"** section recording the gap and naming all six defects, so the design work is captured rather than lost. Diff is smaller, `--help` no longer advertises the flag, and the script lane is back to 766 tests, 0 failures.

### The two documentation findings

**Roadmap registry** — you're right that per-release and per-quarter numbers in a supporting plan create roadmap truth outside `docs/roadmap/data/`. I've marked them **proposed** and stated that adopting them means adding a registry entry. I haven't added one myself: minting a tracked delivery commitment with a wave, lane and priority is a planning decision, not a cleanup side effect.

**dev-fixtures** — verified and you're right, and it matters more than a citation slip. `vite.config.ts:6` imports `resolveDevFixture` *statically*, and the config is evaluated by `vite build`; the preview proxy calls it at line 80. So "dev-only, dynamically imported behind `import.meta.env.DEV`" is not an accurate description of the import graph — and that was the load-bearing half of the argument for excluding 6,673 lines from a guardrail.

Worth flagging that this error originates in the audit itself (#2620: *"Dev fixtures are correctly gated behind `import.meta.env.DEV` with dynamic import… No issue filed"*), and #2619 inherited it, and I repeated it. The plan now states what is actually true: the fixtures may well be absent from the emitted bundle, but the exclusion should be argued on effort grounds with the static config import accepted or isolated — not on a dev-only claim that doesn't hold.

I'll open a separate issue for the tightening mechanism so the six defects and the seam analysis survive as design input.

---

rodoHasArrived:
Fixed in `58f479e0`. The finding is correct, and I reproduced it before changing anything — a file 40 lines over its cap, with a genuinely near-cap file alongside it:

```
- GREW past cap: src/grown.cs now 70 lines (baseline cap 30, exceeded by 40). ...
TIGHT: 2 of 2 tracked file(s) are within 25 line(s) of their cap. 1 of them sit at their cap, where a single added line fails this check.
- src/grown.cs: 70/30 (-40 line(s) spare)
- src/near.cs: 28/30 (2 line(s) spare)
```

Three false statements in one block: the file is not within 25 lines of its cap, it does not sit at its cap, and "a single added line fails this check" describes a file that is already failing. The `-40 line(s) spare` is the visible symptom; the sort order is the one that actually costs information, since ascending headroom puts every over-cap file ahead of the near-cap files the warning exists to surface, and only ten are printed.

This is not a rare path. `_report_trend` runs on failure too, and on failure there is at least one over-cap file by construction — so the contradiction appeared on *every* ratchet failure caused by a grown file.

Bounding the comparison at zero fixes all of it, and loses nothing: over-cap files are already reported by name with exact overage in the failure block immediately above. `pinned` also drops from `headroom <= 0` to `headroom == 0`, which is now the set its wording actually describes. Same output after the fix:

```
TIGHT: 1 of 2 tracked file(s) are within 25 line(s) of their cap. Decompose before extending.
- src/near.cs: 28/30 (2 line(s) spare)
```

Covered by `test_does_not_list_an_over_cap_file_as_tight`, which asserts the over-cap file is absent, the near-cap file is present, and the pinned sentence does not appear. Script lane: 767 tests, 0 failures. Ratchet against the real tree is unchanged (50/50 at or under cap, so nothing in the passing path moves).

---

Issue #2649, opened 2026-08-11
[PRD-002] Eliminate plaintext provider credential persistence and unify credential ownership


rodoHasArrived:
## Tracker facts

| Field | Value |
| --- | --- |
| Row ID | `PRD-002` |
| Priority tier | **P0** — Release blocker |
| Owner | Platform Security + Data Integration |
| Source | [`docs/product/implementation-todo-list.md`](docs/product/implementation-todo-list.md) line 101 |

> No supported production release while this row is open.

## TODO

**Eliminate plaintext provider credential persistence and unify credential ownership.**

Migrate every consumer from the Core plaintext JSON store to the encrypted/audited vault, scope secrets to tenant/connection/account/environment, protect non-Windows vault keys with an OS keyring or KMS-backed mechanism, rotate safely, and securely remove the legacy sidecar.

## Completion evidence required

- [ ] No supported runtime writes `provider-credentials.json`.
- [ ] Migration/rollback/rotation/two-account-isolation tests pass.
- [ ] Raw secrets never appear in files, logs, or APIs.

### Progress

The workstation adapter now delegates to `DataIntegration/Credentials/FileProviderCredentialStore.cs`; it migrates the legacy sidecar only after encrypted persistence succeeds and then fsync-overwrites/deletes it.

## Source of truth

This issue mirrors row `PRD-002` of the [Production-Readiness Master TODO](docs/product/implementation-todo-list.md), named by [`docs/roadmap/data/program-state.yml`](docs/roadmap/data/program-state.yml) as the canonical production-readiness tracker. **The tracker stays authoritative.**

rodoHasArrived:
PR #2677 merged (`d6c3505e`). Recording what that covers against this row's evidence, and what it does not — this issue stays open.

**What landed.** On non-Windows hosts the vault key was created at the umask default (`0644`) in the same directory as the vault it decrypts, so any local account able to read the ciphertext could read the key beside it. `FileAttributes.Hidden`, the only narrowing attempted, is a no-op on Unix. The key is now created `0600` via `FileStreamOptions.UnixCreateMode` — at `open(2)`, not chmod'ed afterwards, since a key written first and narrowed second is readable for the width of the write. A key left over from an earlier release is tightened on next read and logged as disclosed-and-should-be-rotated; a `.mdc` directory this store creates is `0700`.

**Against the completion evidence:**

[...]

Issue #2676, opened 2026-08-12
Paper fill can land outside the observed envelope: PaperMarketObservation.Capture reads quote and trade non-atomically


rodoHasArrived:
`main` is currently red. `verify-dotnet` failed on `304bb8df`, taking `quality-gate` with it, and because Meridian CI evaluates the merge commit this reddens every open PR once CI resumes.

The failure is one test:

```
Meridian.Tests.Execution.PaperFillEnvelopeRegressionTests.AdapterGateway_RestingStop_TriggersOnLaterTrade_NeverBefore

Expected update.AverageFillPrice!.Value to be between 505M and 505.5M
because the triggered stop fills inside the quote envelope in effect at trigger time,
but found 500.5M.
  tests/Meridian.Tests/Execution/PaperFillEnvelopeRegressionTests.cs:247
```

Run: https://github.com/rodoHasArrived/Meridian-main/actions/runs/31540906990/job/93942822334

## This is a product defect, not a flaky assertion

`500.5` is not an arbitrary wrong number — it is exactly the **ask of the pre-update quote**. The test opens with `bid 500 / ask 500.5`, then moves the quote to `bid 505 / ask 505.5` before the trade that triggers the stop. A fill at `500.5` means the order matched a triggered stop against a quote that was no longer in effect.

That is the precise invariant the W9-PAPER-003 suite exists to prove, per its own file header:

> no paper fill can occur at a price outside the observed market-data envelope for the bar or tick in effect

### Mechanism

`PaperMarketObservation.Capture` samples the quote and the trade as two independent reads with nothing holding them together (`src/Meridian.Execution/PaperMatching/PaperMarketObservation.cs:66` and `:72`):

```csharp
if (feed.GetLastQuote(symbol) is { } quote)   // line 66
{
    bid = Positive(quote.BidPrice);
    ask = Positive(quote.AskPrice);
}

decimal? lastTrade = feed.GetLastTrade(symbol) is { } trade ? Positive(trade.Price) : null;  // line 72
```

`LiveMarketDataCache` backs each of these with a separate dictionary and each write is an independent store (`RecordQuote` → `_lastQuotes[symbol] = quote`, `RecordTrade` → `_lastTrades[symbol] = trade`). There is no version, sequence check, or lock spanning the pair, so a reader can observe the quote from before a market update and the trade from after it.

[...]

PR #2677, opened 2026-08-12
Create the provider credential vault key with owner-only permissions

rodoHasArrived:
## Summary

On non-Windows hosts the provider credential vault is AES-GCM encrypted under a 256-bit key that `FileProviderCredentialStore` writes to `.mdc/provider-credentials.key` — the same directory as `provider-credentials.vault`, which that key decrypts.

The key was written through `AtomicFileWriter`, which applies the destination's permissions **only when the destination already exists** (`if (destinationExists) CopySecurityMetadata(...)`). A key file has no destination on first write, so it landed at the umask default — `0644` on a typical Linux host. The only narrowing attempted was `FileAttributes.Hidden`, which is a no-op on Unix.

Any local account that could read the vault could therefore also read the key sitting beside it. Encrypting the vault bought nothing against that reader.

Three changes:

- **`AtomicFileWriter`** gains a `byte[]` overload taking an explicit `UnixFileMode`, applied via `FileStreamOptions.UnixCreateMode` so the mode reaches `open(2)` rather than being chmod'ed afterwards — a key written first and narrowed second is readable for the width of the write, and the temp file this method renames holds the real bytes. Passing a mode also suppresses the copy of the destination's security metadata; without that, replacing an over-permissive key would restore the very mode being corrected. The existing three-argument overload delegates with no mode and is behaviourally unchanged.
- **`FileProviderCredentialStore`** creates the key `0600`, and on reading a key that predates this change tightens it and warns that the key should be treated as disclosed and credentials rotated. Repair rather than refuse is deliberate — failing closed would brick every Linux install that already has a `0644` key, a worse outcome than the exposure it would announce.
- **A `.mdc` directory this store creates** is now `0700`. An existing one is left as the deployment configured it: it also holds credential state written by other services, and silently removing access a running install depends on is not a change to make as a side effect. The key file's own mode is what protects the key, and that is enforced either way.

The vault, backup, and audit files keep their current permissions. The vault is ciphertext and the audit holds only masked previews, so their exposure was bounded by the key's — which is the thing being fixed.

## Reason

Advances #2649 (PRD-002, P0 release-blocker), whose scope explicitly includes protecting non-Windows vault keys. **It does not close the issue** — the tracker also requires an OS keyring or KMS-backed mechanism, which is a design decision rather than a permissions fix, plus rotation and two-account-isolation evidence. Filed against the tracker row rather than as its own issue because it is squarely inside that row's stated scope.

One related exposure found and deliberately **not** fixed here: `OAuthTokenRefreshService` writes `.mdc/oauth_tokens.json` and `CredentialTestingService` writes `.mdc/credential_status.json`, both at the umask default. Those are a different service's files; folding them in would widen this change past the defect it describes. Worth a follow-up.

## Testing performed

Six tests added, all guarded to skip on Windows where DPAPI applies and no key file exists:

`AtomicFileWriterTests` — the requested mode is applied on create; an explicit mode overrides an existing wider mode (the regression that would silently re-widen a corrected key); and omitting the mode still preserves existing permissions, so the current callers' behaviour is pinned.

`ProviderCredentialStoreTests` — `SaveAsync` creates the key owner-only; `SaveAsync` creates `.mdc` owner-only; and a key left world-readable by an earlier release is tightened on next read **and still decrypts**, so the repair cannot cost access to the vault.

`dotnet` is not available in this container, so the .NET lane was run on a hosted runner via the `Targeted Test` workflow rather than locally. Local gates run here: file-size ratchet OK, `git diff --check` clean.

## Safety review

## Governance changes

---

### DIFF
--- a/src/Meridian.DataIntegration/Credentials/FileProviderCredentialStore.cs
+++ b/src/Meridian.DataIntegration/Credentials/FileProviderCredentialStore.cs
@@ -605,7 +605,7 @@ public sealed class FileProviderCredentialStore : IProviderCredentialStore
 
     private async Task WriteVaultAsync(ProviderCredentialVault vault, CancellationToken ct)
     {
-        Directory.CreateDirectory(_directoryPath);
+        EnsureVaultDirectory();
         vault.Version = VaultVersion;
         vault.UpdatedAt = DateTimeOffset.UtcNow;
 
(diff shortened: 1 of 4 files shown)


rodoHasArrived:
`verify-dotnet` failed here on a test this change cannot reach. Evidence rather than assertion:

**The failure**

```
Meridian.Tests.Execution.PaperSessionRecoveryConcurrencyTests.Scenario_CancelledHydration_NextInitialisationAttemptCanRetry
System.Threading.Tasks.TaskCanceledException : A task was canceled.
  at CancelFirstLoadStore.LoadAllSessionsAsync(ct)          PaperSessionRecoveryConcurrencyTests.cs:991
  at PaperSessionPersistenceService.InitialiseCoreAsync(..) PaperSessionPersistenceService.cs:103
  at PaperSessionPersistenceService.DisposeAsync()          PaperSessionPersistenceService.cs:1302
```

**Why it is not this diff**

- `PaperSessionPersistenceService` contains zero references to `AtomicFileWriter` — the only production type this PR touches outside `Meridian.DataIntegration`.
- The test drives `CancelFirstLoadStore : MemoryPaperSessionStore`, an in-memory stub. There is no file I/O anywhere in its path, so a file-permission change has nothing to act on.
- The `AtomicFileWriter` change is additive. The pre-existing three-argument overload delegates with `unixCreateMode: null`, and every mode-dependent branch is guarded on non-null, so existing callers execute the identical path. `WriteAsync_Bytes_WithoutModeStillPreservesExistingPermissions` pins that.
- The exception escapes from `DisposeAsync()`, i.e. teardown after the assertions already passed.

**Confirmed by re-running the same commit**

| Run | Commit | Scope | Result |
| --- | --- | --- | --- |
| Meridian CI `verify-dotnet` | `d6bb9d45` | `core-execution-strategy` shard | failed (1 of 1,939) |
| [Targeted Test 31563585625](https://github.com/rodoHasArrived/Meridian-main/actions/runs/31563585625) | `d6bb9d45` | `FullyQualifiedName~Meridian.Tests.Execution` | **passed** |

Same tree, same runner image, opposite outcomes — a race, not a regression.

This is the second intermittent failure in the `core-execution-strategy` shard tonight from a different test; the first reddened `main` itself and is tracked as #2676. I've added this one there, because two independent teardown/cancellation races in one shard reads like a shard-level problem rather than two coincidences.

**On this PR's own tests:** [Targeted Test 31562809431](https://github.com/rodoHasArrived/Meridian-main/actions/runs/31562809431) ran `AtomicFileWriterTests` and `Application.Config.ProviderCredentialStoreTests` against `d6bb9d45` and passed, so the six new tests compile and hold. `dotnet` is unavailable in my environment, which is why verification here runs on hosted runners.

No fix pushed for this failure, deliberately — there is nothing in this diff to fix, and re-running is what clears it.

---

PR #2679, opened 2026-08-12
Write persisted OAuth tokens owner-only

rodoHasArrived:
## Summary

`.mdc/oauth_tokens.json` holds OAuth **access and refresh tokens in cleartext**, in the same directory as the provider credential vault, and was written at the umask default — `0644` on a typical Linux host. Any local account could read live bearer credentials for every connected provider.

This is the exposure I found while fixing the vault key in #2677 and deliberately scoped out of that PR as belonging to a different service. The `AtomicFileWriter` seam #2677 added is what makes this a small change, so it's the natural follow-on.

Two differences from the vault key make this **worse**, not equivalent:

- The vault is ciphertext, so its key file was the thing needing protection. Here there is no encryption at all — the file mode is the entirety of the protection.
- A vault key is only useful alongside the vault it opens. An access token is useful on its own, to anyone, against a live third-party API.

Changes:

- **`AtomicFileWriter`** gains the text counterpart of the `byte[]` mode overload from #2677, same contract: applied at creation via `FileStreamOptions.UnixCreateMode` rather than chmod'ed afterwards, and suppressing the destination's security-metadata copy so replacing an over-permissive file cannot restore its mode. The existing three-argument overload delegates with no mode and is unchanged.
- **`OAuthTokenRefreshService`** writes the token file `0600`, and tightens it **on load** as well as on write. Rewriting only happens on the next refresh — hours away, or never while the tokens remain valid — so repairing on write alone would leave an existing install exposed indefinitely. The warning names the tokens as disclosed and says to revoke them, because unlike a local vault key they can be replayed remotely.

**Deliberately not changed:** `credential_status.json`, written to the same directory by `CredentialTestingService`. `StoredCredentialStatus` carries a provider name, timestamps, a status enum, and a failure count — no secret material. Tightening it would be applying the pattern rather than fixing an exposure.

**Known inconsistency this does not resolve:** whichever service creates `.mdc` first decides its mode, and only `FileProviderCredentialStore` narrows it. The file's own mode protects the tokens either way, but that directory would be better with a single owner than with two services each deciding.

## Reason

Advances #2649 (PRD-002, P0 release-blocker), whose scope covers credential persistence. **Does not close it** — the OS keyring / KMS mechanism, rotation, and two-account isolation all remain, as recorded in the issue.

## Testing performed

Six tests, all guarded to skip on Windows where ACLs rather than mode bits govern:

`OAuthTokenPersistencePermissionTests` (new) — `StoreTokenAsync` writes the file owner-only; a file left world-readable by an earlier release is tightened at construction **and the tokens still load**; and rewriting an already-tightened file keeps it owner-only.

`AtomicFileWriterTests` — the text overload applies the requested mode; an explicit mode overrides an existing wider one; and it still writes UTF-8 **without a BOM** when a mode is given, since supplying a mode routes through a byte path and a BOM there would corrupt readers that token-split raw content.

`dotnet` is unavailable in this container, so the .NET lane runs on a hosted runner via `Targeted Test`. Local gates run here: file-size ratchet OK, `git diff --check` clean, source-README validation clean, and generated docs regenerated to convergence.

## Safety review

## Governance changes

---

### DIFF
--- a/docs/generated/repository-structure.md
+++ b/docs/generated/repository-structure.md
@@ -8750,6 +8750,7 @@ Meridian-main
 │   │   │   │   ├── ConfigSchemaIntegrationTests.cs
 │   │   │   │   ├── ConfigurationUnificationTests.cs
 │   │   │   │   ├── ConfigValidatorCliTests.cs
+│   │   │   │   ├── OAuthTokenPersistencePermissionTests.cs
 │   │   │   │   ├── ProviderCredentialResolverTests.cs
 │   │   │   │   └── ProviderCredentialStoreTests.cs
 │   │   │   ├── Coordination
--- a/docs/status/doc-health-dashboard.json
+++ b/docs/status/doc-health-dashboard.json
@@ -1,6 +1,6 @@
 {
   "total_files": 643,
-  "total_lines": 124865,
+  "total_lines": 124866,
   "orphaned_count": 255,
   "no_heading_count": 148,
   "stale_count": 0,
@@ -2594,7 +2594,7 @@
     },
     {
       "path": "docs/generated/repository-structure.md",
-      "line_count": 10395,
+      "line_count": 10396,
       "has_heading": true,
       "todo_count": 8,
       "last_modified_utc": "1970-01-01T00:00:00+00:00",
--- a/docs/status/doc-health-dashboard.md
+++ b/docs/status/doc-health-dashboard.md
@@ -20,7 +20,7 @@ Data sources: `repo markdown (*.md)`, `file modification metadata`
 | Metric | Value |
 | -------- | ------- |
 | Total documentation files | 643 |
-| Total lines | 124,865 |
+| Total lines | 124,866 |
 | Average file size (lines) | 194.2 |
 | Orphaned files | 255 |
 | Files without headings | 148 |
--- a/src/Meridian.Application/Config/Credentials/OAuthTokenRefreshService.cs
+++ b/src/Meridian.Application/Config/Credentials/OAuthTokenRefreshService.cs
@@ -326,12 +326,47 @@ public sealed class OAuthTokenRefreshService : IAsyncDisposable
         return TokenStatus.Valid;
     }
 
+    private const UnixFileMode OwnerOnlyFileMode = UnixFileMode.UserRead | UnixFileMode.UserWrite;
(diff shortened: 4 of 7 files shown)


rodoHasArrived:
Two CI rounds here, one my fault and one not. Separating them.

**Mine.** The first hosted run failed to compile: I used `NotStartWith` on a byte collection, which FluentAssertions does not have. Fixed in `2368b43b` with a byte-exact comparison against a BOM-free encode — the stronger assertion anyway, since it pins both the absence of a BOM and the content, where a prefix check only ruled out the three bytes it named. `dotnet` is unavailable in my environment, which is exactly why the change goes through a hosted run before I claim anything about it.

**Not mine.** `verify-dotnet` then failed on `SecurityMasterWorkbenchOptionsBindingTests`, in the `core-ui-other` shard. The build reported `0 Error(s)`, and every other shard passed — including `core-execution-strategy` at 1,939/1,939, and `core-application` where this PR's new tests actually live.

That failure is env-var pollution between parallel xUnit collections, filed as #2680 with the full trace. Briefly: the failing test sets no environment at all, but `AddWorkstationSharedServices` decides "production" from `ProductionServiceRegistrationPolicy.IsProductionEnvironment()`, which reads process-global `ASPNETCORE_ENVIRONMENT`. A sibling class sets precisely that:

```csharp
// tests/Meridian.Tests/Ui/ProductionStartupPolicySmokeTests.cs:13
using var environment = new EnvironmentVariableScope("ASPNETCORE_ENVIRONMENT", "Production");
```

Neither class carries a `[Collection]` attribute, so xUnit runs them concurrently, and `SetEnvironmentVariable` mutates the whole process. The scope restores correctly — the bug is that a correctly-scoped process-global is still global to every thread while it is held.

This PR cannot have caused it. It changes `AtomicFileWriter` and `OAuthTokenRefreshService`, neither of which is in that path, and its new test class is in `core-application` — a separate `dotnet test` invocation, a separate process — so it could not have perturbed that run's scheduling either.

Re-run queued, which is the only remedy available today. That is the third distinct flaky family this week (#2676 covers two more, both in `core-execution-strategy`), and #2680 argues the real fix is to stop reading ambient posture deep in the registration graph rather than to keep adding collection attributes.

---

Issue #2614, opened 2026-08-11
No shared string primitives: `NormalizeOptional` declared in 62 files, `RequireText` in 31, `Normalize` in 26


rodoHasArrived:
Found during a codebase-wide audit (duplication / pragmatism).

## Summary

The codebase has no shared string/normalization primitives module. The same handful of trivial helpers is redeclared as `private static` in hundreds of files. This is low-risk individually but it is the single largest source of mechanical duplication in `src/`, and it is what makes the higher-severity drift bugs (#2610, #2611) possible in the first place.

## Evidence

Distinct files declaring their own private/internal static copy:

| Helper | Files |
|---|---|
| `NormalizeOptional` | **62** |
| `RequireText` | 31 |
| `Normalize` | 26 |
| `NormalizeCurrency` | 19 |
| `IsSha256` | 18 |
| `ComputeSha256` | 18 |
| `FirstNonBlank` | 15 |
| `EnsureHumanOrigin` | 15 |
| `FormatBytes` | 12 |
| `NormalizeFundProfileId` | 12 |
| `FormatDecimal` | 11 |
| `ParseDecimal` | 10 |
| `FormatDuration` | 10 |
| `NormalizeSymbol` | 10 |
| `NormalizeRequired` | 10 |
| `ValidateIdentifier` | 10 |
| `TrimOrNull` | 9 |
| `ParseDate` | 9 |
| `FormatCurrency` | 9 |

`NormalizeOptional` already exists in at least two spellings that are behaviorally equivalent but textually distinct:

```csharp
// 20 files
=> string.IsNullOrWhiteSpace(value) ? null : value.Trim();

// 6 files
{ var normalized = value?.Trim(); return string.IsNullOrWhiteSpace(normalized) ? null : normalized; }
```

…plus a family of near-name variants that fragment the concept further: `NormalizeOptionalValue`, `NormalizeOptionalToken`, `NormalizeOptionalDimensionValue`, `NormalizeOptionalTenantTemplateId` (declared identically in two separate files), `NormalizeOptionalScope`, `NormalizeOptionalIdentity`.

## Why it matters

The direct cost is modest — these are one-liners, and the current copies of `NormalizeOptional` do agree. The real cost is structural:

[...]

Issue #2610, opened 2026-08-11
SHA-256 digest casing is inconsistent: uppercase producers exist and some validators reject uppercase outright


rodoHasArrived:
> **Revised after deeper verification.** The original version asserted a demonstrated live failure in a specific code path. I traced the flows and could **not** prove an uppercase producer currently reaches a lowercase-only gate at runtime, so that claim has been softened. In exchange, verification turned up substantially stronger evidence — confirmed uppercase digest producers and a validator that explicitly rejects uppercase — which is now the core of the issue.

Found during a codebase-wide audit (robustness / duplication).

## Summary

`Convert.ToHexString` returns **uppercase** hex in .NET. Most digest sites in this codebase append `.ToLowerInvariant()`; **three do not**. Meanwhile `IsSha256` is redeclared in 18 files across two incompatible families — some accept uppercase, some reject it — and one validator throws an explicit *"must be a lowercase SHA-256 value"*. There is no shared digest primitive and no test coverage for any of it.

### 1. Confirmed uppercase producers

| Location | Code |
|---|---|
| `src/Meridian.Reporting/CertifiedReportingSnapshotBuilder.cs:220` | `return Convert.ToHexString(SHA256.HashData(stream.ToArray()));` |
| `src/Meridian.Reporting/ReportingOrchestrationService.cs:1627` | `return Convert.ToHexString(bytes);` |
| `src/Meridian.Strategies/Services/ReconciliationCaseWorkflowService.cs:63` | `return Convert.ToHexString(bytes);` |

The first one is the clearest sign this is an oversight rather than a decision. In `CertifiedReportingSnapshotBuilder.Build`, the same `hash` value is used twice on adjacent lines — lowercased for one field, left uppercase for the other:

```csharp
var receipt = new ReportingSnapshotReceipt(
    SnapshotId: $"reporting-snapshot-{hash[..24].ToLowerInvariant()}",   // lowercased
    ContentHash: hash,                                                   // left UPPERCASE
    ...
```

### 2. Validators disagree about whether uppercase is legal

`IsSha256` is declared in **18 files**, in two families:

| Variant | Copies | Uppercase? |
|---|---:|---|
| `value.All(Uri.IsHexDigit)` | 15 | **accepted** |
| `value.All(c => c is >= '0' and <= '9' or >= 'a' and <= 'f')` | 4 | **rejected** |

Plus a third spelling that adds an explicit lowercase assertion and a matching error message:

```csharp
[...]

PR #2690, opened 2026-08-15
Route ComputeSha256 through the canonical Sha256Digest primitive

rodoHasArrived:
## Summary

Next tranche of #2614. Removes the duplicated SHA-256 hashing behind `ComputeSha256` — 20 declarations across 18 files — by routing it to the `Sha256Digest` primitive that #2610 already built.

## `IsSha256` is already done, and checking that first changed this change

The issue lists `IsSha256` at 18 files and warns it had *already* split into two incompatible families. That was the reason to take it first — but it landed with #2610 as PR #2637. `Sha256Digest` owns the predicate now, and the three surviving `IsSha256` declarations are thin delegations that add real semantics:

| File | Adds |
| --- | --- |
| `AccountingPostingCandidateService.cs` | `.Trim()` before validating |
| `AssetAccountingEventSpineService.cs` | rejects an all-zero digest |
| `AssetAccountingEventDtos.cs` | rejects an all-zero digest |

Those are deliberate and stay. Verifying it before starting is what redirected this to the half that was actually left.

## What this does

`Sha256Digest` already exposes exactly the two shapes needed — `Compute(ReadOnlySpan<byte>)` and `ComputeUtf8(string)` — and all 20 `ComputeSha256` declarations were behaviourally identical to them: SHA-256 → hex → lowercase. **No divergence this time**; `Convert.ToHexStringLower` and `Convert.ToHexString(...).ToLowerInvariant()` produce the same string, so unlike `NormalizeOptional` in #2688 there was no latent behaviour split to preserve.

- **16 files with a single overload** — call sites rewritten to `Sha256Digest.Compute` / `.ComputeUtf8`, private copy deleted. This is the pattern #2637 established.
- **2 files declaring both overloads** (`LedgerScheduledReportExportPackageBuilder`, `PostgresReportingRunStore`) — a name-only rewrite cannot tell which overload a call meant, so these keep their overloads as one-line delegations. The duplicated hashing is gone either way, which is the part that could drift.

## Reason

Partial fix for #2614; not closing it. Remaining: `Normalize` (26), `NormalizeCurrency` (19), `FormatBytes` (12), and the domain-owned canonicalisers.

## A finding worth separating from this PR

Grepping for the **hashing itself** rather than for helper names finds **275 inline `SHA256.HashData` sites** outside the primitive, against the 18 the helper-name audit found. Concentrated in `Meridian.Ui.Shared` and `Meridian.Strategies`:

```
9  StatementReconciliationReportWorkflowService.cs
8  ReportingAccessGrantService.cs
7  FileEvidenceArtifactStore.cs
6  ReportPackDeliveryService.cs
5  ReportingGovernanceCanonicalValidation.cs
```

Consolidating named helpers does not touch any of it. That is a far larger surface than #2614 describes, and it is where the next `IsSha256`-style divergence would hide — an open-coded hash has no name to audit. Flagging rather than folding it in; happy to file it separately.

## Testing performed

- [ ] Relevant unit tests were added or updated — not applicable; `Sha256DigestTests` already covers the primitive (added by #2637), and this only redirects callers to it
- [ ] `bash scripts/ci.sh` completed successfully — **not run; no .NET SDK in this environment**
- [ ] GitHub Actions `quality-gate` passed — pending
- [ ] Relevant integration tests were added or updated — not applicable

> **Not compiled locally.** Checked by inspection instead:
> - **0 orphaned callers.** Reaching that required fixing three partial-class siblings whose declaration lived in another file — the same hazard as #2688, and the reason a name-based sweep needs a caller check afterwards.
> - **Brace balance unchanged from HEAD** in every touched file.
> - Both existing ratchets exit `0`; docs generators idempotent.

`ProviderLedgerReconciliationService.cs` would have gone one line over its cap from the added `using`, and has two call sites, so those are fully qualified and the import dropped. Same trade as the previous tranche, where a one-call-site file got qualified and a 48-call-site file got a justified cap bump.

## Safety review

## Governance changes

---

### DIFF
--- a/src/Meridian.Ledger/LedgerReportPackBuilder.cs
+++ b/src/Meridian.Ledger/LedgerReportPackBuilder.cs
@@ -4,6 +4,7 @@ using System.Text;
 using System.Text.Json;
 
 using static Meridian.Contracts.Ledger.LedgerCurrencyRounding;
+using Meridian.Contracts.Integrity;
 
 namespace Meridian.Ledger;
 
@@ -53,7 +54,7 @@ public static class LedgerReportPackBuilder
 
         var signature = new LedgerReportPackSignature(
             "SHA256",
-            ComputeSha256(payload),
+            Sha256Digest.ComputeUtf8(payload),
             request.GeneratedBy,
             request.GeneratedAtUtc);
 
@@ -170,13 +171,13 @@ public static class LedgerReportPackBuilder
         }
 
         var content = builder.ToString();
-        return new LedgerReportPackArtifact("tax-lot-realized-gains.csv", "text/csv", content, ComputeSha256(content));
+        return new LedgerReportPackArtifact("tax-lot-realized-gains.csv", "text/csv", content, Sha256Digest.ComputeUtf8(content));
     }
 
     private static LedgerReportPackArtifact CreateCsvArtifact(string name, IReadOnlyList<LedgerChartBalance> rows)
     {
         var content = BuildCsv(rows);
-        return new LedgerReportPackArtifact(name, "text/csv", content, ComputeSha256(content));
+        return new LedgerReportPackArtifact(name, "text/csv", content, Sha256Digest.ComputeUtf8(content));
     }
 
     private static LedgerReportPackArtifact CreateCashFlowArtifact(string name, LedgerCashFlowStatement? cashFlow)
@@ -207,7 +208,7 @@ public static class LedgerReportPackBuilder
         }
 
         var content = builder.ToString();
-        return new LedgerReportPackArtifact(name, "text/csv", content, ComputeSha256(content));
+        return new LedgerReportPackArtifact(name, "text/csv", content, Sha256Digest.ComputeUtf8(content));
     }
 
     private static void AppendCashFlowTotal(StringBuilder builder, string label, decimal amount)
@@ -273,7 +274,7 @@ public static class LedgerReportPackBuilder
         }
 
         var content = builder.ToString();
-        return new LedgerReportPackArtifact(name, "text/csv", content, ComputeSha256(content));
+        return new LedgerReportPackArtifact(name, "text/csv", content, Sha256Digest.ComputeUtf8(content));
     }
 
     private static LedgerReportPackArtifact CreateFinancialStatementsJsonArtifact(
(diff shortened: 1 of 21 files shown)


rodoHasArrived:
CodeQL reports **3 high** alerts under "New alerts in code changed by this pull request". I can't read them yet — the check run's `output.text` is empty and no `github-advanced-security` review threads have posted, so the detail is only on the security tab. Recording what I have verified rather than waiting silently, and I'll act on the specifics when they land.

**The diff introduces no new hashing behaviour, and that is checkable rather than asserted.** Filtering the C# diff down to lines that are not a redirect, an import, or a comment leaves only deletions — every removed line is a copy of:

```csharp
Convert.ToHexString(SHA256.HashData(...)).ToLowerInvariant()
```

and the single algorithmic addition is `Sha256Digest.Compute(Encoding.UTF8.GetBytes(payload))`, which preserves the identical UTF-8 → SHA-256 → lowercase-hex path. `Sha256Digest.Compute` uses `Convert.ToHexStringLower`, producing the same string; `ComputeUtf8` adds an `ArgumentNullException.ThrowIfNull` guard, which is stricter than what it replaced, not looser. No algorithm, encoding, casing, or input path changed anywhere in this PR.

So whatever these three are, they are not new weak crypto introduced here. Two readings remain, and I am **not** assuming which:

1. **Reattribution.** CodeQL files alerts against changed lines, so a pre-existing pattern on a line this PR touched is reported as new — the same thing that happened on #2671, where I initially called such alerts "pre-existing" and had to correct that they are new *alerts* even when not a new *vulnerability*.
2. **Newly traceable dataflow.** Consolidating twenty local helpers into one primitive gives CodeQL a single modelled sink reachable from more sources, so a finding that was always true may only now be visible. If that is what happened, the finding is real and predates this change — and surfacing it is the consolidation doing its job, exactly as with the governance predicate on #2671.

Reading 2 would make these worth acting on rather than waving through, which is why I am not calling them benign. I'll read the alerts once they are retrievable and either fix them here or explain why they belong elsewhere.

The accompanying `2 configurations not found` warning (`/language:actions`, `/language:python`) is unrelated to the diff — those configs exist on `main` but not on the PR ref.

---

Issue #2618, opened 2026-08-11
Endpoint error handling is inconsistent: `EndpointHelpers` used in 46 of 143 endpoint files


rodoHasArrived:
Found during a codebase-wide audit (pragmatism / consistency).

## Summary

`EndpointHelpers` exists specifically to standardize fail-closed service checks, cancellation propagation, and error-to-`ProblemDetails` mapping across workstation endpoints. It is well designed and well documented — but it is used in fewer than a third of endpoint files, so the guarantees it encodes are not uniform across the API surface.

## Evidence

- `src/Meridian.Ui.Shared/Endpoints/` — **143** files, **1,158** `.MapGet/MapPost/MapPut/MapDelete/MapPatch` registrations
- Files referencing `EndpointHelpers.` — **46**
- Remaining hand-rolled `catch (Exception` in endpoint files — **62**

Highest concentrations of hand-rolled handling:

| File | Raw `catch (Exception` |
|---|---:|
| `FundStructureEndpoints.cs` | 12 |
| `WorkstationEndpoints.ReportingAuthority.cs` | 5 |
| `HistoricalEndpoints.cs` | 4 |
| `WorkstationEndpoints.StatementConnectors.cs` | 3 |
| `StrategyLifecycleEndpoints.cs` | 3 |
| `FailoverEndpoints.cs` | 3 |
| `BackfillScheduleEndpoints.cs` | 3 |

## Why it matters

The helper encodes three behaviors that are easy to get wrong by hand, and it documents why each one matters:

```csharp
// src/Meridian.Ui.Shared/Endpoints/EndpointHelpers.cs
if (service is null)
    return ApiProblemDetails.ServiceUnavailable(context, typeof(TService).Name);   // fail-closed
...
catch (OperationCanceledException) when (ct.IsCancellationRequested)
{
    // The request was aborted; propagate so the host treats it as a cancellation
    // instead of reporting an error response to a client that is no longer listening.
    throw;
}
catch (Exception ex) { return FormatErrorResult(ex, context); }                     // uniform ProblemDetails
```

Endpoints that hand-roll this have to independently remember to re-throw `OperationCanceledException`. Getting that wrong converts a client disconnect into a logged 500 — noise that masks real failures, and a direct conflict with the repo guardrail to *keep cancellation flow intact for async code* (CLAUDE.md). Inconsistent handling also means the error payload shape varies by endpoint, so clients cannot rely on one contract.

[...]

PR #2699, opened 2026-08-15
Stop three endpoint handlers from swallowing a client disconnect (#2618, partial)

rodoHasArrived:
## Summary

Three endpoint handlers could catch an `OperationCanceledException` raised by a client disconnect and answer as though a normal error had occurred. Each now re-throws cancellation ahead of its generic handler, matching `EndpointHelpers.HandleAsync`'s token overload. A new guard plus test keeps the endpoint surface at zero.

Behaviour for every non-cancellation error is unchanged.

This is a **partial** fix for #2618 and should not close it — the consolidation lane (step 2) is untouched.

## Reason

#2618's step 1 asks which of the 62 hand-rolled `catch (Exception` sites miss the cancellation re-throw — "the one with a real runtime consequence, and the cheapest thing to get wrong." So I classified all 62 before changing anything:

| | Count | Why |
| --- | ---: | --- |
| **Filtered** | 35 | `when (ex is ArgumentException or …)` / `when (ex is not OperationCanceledException)` — a filter already lets cancellation past |
| **Guarded** | 20 | An earlier clause in the same chain catches `OperationCanceledException`, whether it re-throws or answers 499 |
| **Bare but synchronous** | 4 | No token flows in, so no disconnect can land in the catch. Untidy, not a defect. |
| **Bare *and* reachable** | **3** | A cancellation token reaches an awaited call |

The headline number is 3, not 62. The endpoint surface is mostly already correct — but the three exceptions each do something distinctly wrong with a hang-up:

- **`CheckpointEndpoints.cs:241`** — answers `400 Bad Request` with `"Resume failed: The operation was canceled."`, blaming the caller for hanging up.
- **`MessagingEndpoints.cs:236`** — increments `s_totalFailed` and appends to `s_errorLog`. An aborted request puts a webhook delivery **that never happened** into the operator-visible messaging error queue. This is the worst of the three: the damage outlives the request.
- **`MessagingEndpoints.cs:382`** — answers `200 OK` with `retried = false`, which reads as "the retry ran and failed".

The `when (<token>.IsCancellationRequested)` filter is load-bearing at the messaging sites: a webhook whose own HTTP call times out throws `TaskCanceledException` **without** the request being aborted, and that is a genuine delivery failure that must still be counted. An unconditional re-throw would lose it.

I did not rewrite these onto `EndpointHelpers.GuardAsync` — Checkpoint's 400 and Messaging's counter mutation are not shapes it reproduces, and swapping them would change behaviour for real errors. That is the issue's step 2 and belongs in its own change.

## Testing performed

**Not compiled locally — this container has no .NET SDK.** CI is the first real compile; treat the checkboxes accordingly.

`build/scripts/ci/check-endpoint-cancellation.py` generalises the classifier, and `tests/scripts/test_check_endpoint_cancellation.py` enforces it — 12 cases: nine pin the shapes that are and are not defects (filtered catch, preceding re-throw, preceding 499, synchronous handler, `await` with no token, `catch` inside a comment, braces inside a string), and a live-tree case asserts the endpoint surface stays at zero.

That last case is the one that matters, and it is falsifiable: reverting just the three source fixes fails it, naming exactly those three sites.

```
- [('src/Meridian.Ui.Shared/Endpoints/CheckpointEndpoints.cs', 241),
-  ('src/Meridian.Ui.Shared/Endpoints/MessagingEndpoints.cs', 236),
-  ('src/Meridian.Ui.Shared/Endpoints/MessagingEndpoints.cs', 382)]
```

Enforcement rides the existing script-test lane: `run-script-tests.py` auto-discovers `tests/scripts/` and runs inside `verify_workflows()`, which `quality-gate` aggregates. So the guard is live **without** touching `scripts/ci.sh`.

Also run locally: full script lane green at 795 tests, both existing ratchets exit 0, docs regenerated rather than hand-edited.

### Scope note

Running the same scan across all of `src/` reports **278** further sites — 84 `Meridian.Wpf`, 44 `Application`, 41 `Ui.Services`, 40 `Infrastructure`, 29 `Storage`, 15 `Execution`, the rest scattered. Most are UI and background code where dropping a superseded operation is often deliberate, so that is a different question from #2618. The guard stays scoped to the endpoint surface the issue describes rather than silently widening.

## Safety review

## Governance changes

- [ ] `.github/workflows/**`
- [ ] `.github/CODEOWNERS`
- [ ] `AGENTS.md`
- [ ] `scripts/ci.sh`

---


rodoHasArrived:
Six of the seven review findings were right and are fixed in `62cf8a01d`. The seventh I'm declining, with reasoning.

**Fixed**

| Finding | Was | Now |
| --- | --- | --- |
| Qualified catches | Pattern required `Exception` right after the paren | Accepts `System.` / `global::System.` prefixes |
| `TaskCanceledException` | Counted as a chain guard | Doesn't — it derives from `OperationCanceledException`, so it can't catch the base type `ct.ThrowIfCancellationRequested()` throws. Only `OperationCanceledException` counts |
| Opaque filters | Every filtered catch exempted | Only filters that *provably* exclude cancellation; `when (ShouldHandle(ex))` is now reported |
| Token vs. await | `'await' in body and TOKEN.search(body)` — two independent substring checks | Each awaited statement is sliced to its terminator; the token must appear inside that slice |
| Token spelling | Fixed allowlist | Named arguments and arbitrary names (`cancellationToken: requestCancellation`) |
| Missing scan dir | Skipped silently | Raises |

The token-vs-await one was the most worth having — a false positive in a CI gate blocks valid code, which is worse than the false negatives the others describe.

**The filter change found something, then unfound it.** Tightening filter trust initially flagged a fourth site, `WorkstationEndpoints.cs:3769`:

```csharp
catch (Exception) when (!ct.IsCancellationRequested)
```

That is not a defect — it's a different correct idiom. While the token is cancelled the filter is false, so the catch declines entirely and a disconnect always propagates. My stricter rule simply didn't model it, so the finding was right in principle and my implementation of it was too strict. The rule now recognises that shape plus the boolean structure around it: a disjunction is safe only if every branch is safe, a conjunction as soon as one term is.

Fourteen new tests (26 total) pin each fix, including the two shapes that must stay *clean* — `!ct.IsCancellationRequested` and token-not-on-the-awaited-call — so a future tightening can't quietly reintroduce a false positive. Live tree still reports zero; reverting the three source fixes still fails the guard naming exactly those three sites. Script lane green at 809 tests.

---

**Declining: "Preserve evidence of partially completed webhook sends"** (`MessagingEndpoints.cs:242`)

The gap is real but it predates this change, and the change doesn't widen it — it narrows the harm.

Before, a mid-loop disconnect recorded `s_totalFailed++` plus an error entry saying `"The operation was canceled."` That is not partial-delivery evidence either; it asserts the *whole* test send failed, which is equally wrong about a run where webhook 1 succeeded and webhook 2 was cancelled. Neither the old nor the new code records "1 delivered, 1 indeterminate" — capturing that requires per-webhook accounting inside `SendMessageAsync`, not a different catch clause in the endpoint.

On the duplicate-send risk specifically, the change helps rather than hurts: `/api/messaging/errors/{messageId}/retry` retries entries out of `s_errorLog`. Under the old behaviour a cancelled send landed in that log and was retryable — which is precisely how you'd re-send a notification that had already gone out. It no longer does.

Happy to file the per-webhook delivery-evidence gap as its own issue if that's useful; it's a real improvement, just not one this diff should smuggle in.

---

rodoHasArrived:
Round three: nine findings, all technically correct as stated, all false *negatives* rather than false positives. I measured each against the scanned surface before acting, since round two's checkable claims were the ones that turned up real defects. Pushed `e4584fe15`.

**Six describe shapes with zero occurrences** in `src/Meridian.Ui.Shared/Endpoints`: an await inside an interpolated string, two different tokens in one handler, `is not TaskCanceledException`, nested-call exception filters, an explicit `throw new OperationCanceledException`, and `catch (SystemException)`. The entire endpoint surface declares exactly two token names — `ct` and `cancellationToken`.

**Fixed anyway**, where the fix is small and the current code is simply wrong:

- **Token detection** — the allowlist is gone entirely, which is what the round-two comment asked for. Names now come from the file's own `CancellationToken foo` declarations, so `requestToken` and `stoppingToken` are recognised without the pattern anticipating them. Worth noting this is *stricter* than broadening the suffix match: matching `\w*Token` would have started treating `authToken` and `accessToken` as cancellation tokens, introducing a false positive to fix a false negative. There's a test pinning that it doesn't.
- **`catch (SystemException)`** — `OperationCanceledException` derives from it, so it's a catch-all for cancellation.
- **`is not TaskCanceledException`** — was treated as excluding all cancellation. It's a subclass, so a base `OperationCanceledException` still satisfies the filter. Real logic error.
- **Nested filters** — `when (ShouldHandle(ex, Ctx()))` failed to parse, which drops the clause silently rather than reporting it. Two levels now.
- **Explicit throws** — `throw new OperationCanceledException(ct)` joins `ThrowIfCancellationRequested` as a reachability source.

---

**Declining: require the guard clause to rethrow or answer 499 before exempting the chain.**

Eleven `catch (OperationCanceledException)` clauses in scope neither rethrow nor return 499, and most are legitimate:

| Site | Body | Why it's correct |
| --- | --- | --- |
| `StreamEndpointHelpers.cs:68` | heartbeat comment | Cancellation raised by a *heartbeat interval* in an SSE loop — ordinary control flow, not a disconnect |
| `WorkstationEndpoints.Stream.cs:97` | heartbeat comment | Same |
| `StrategyLifecycleEndpoints.cs:102/181/262` | `ToEndpointFailureResult(..., Blocked)` | Deliberately converts cancellation into a structured Blocked outcome |

Implementing this literally would flag exactly the code where catching cancellation is correct.

It also asks a different question than this check does. The guard answers "can cancellation reach the generic catch-all?" An earlier `OperationCanceledException` clause stops it reaching there *whatever its body does*. Whether that clause then handles cancellation well is a worthwhile review question, but it isn't reachability, and folding it in would make the gate mean two things at once.

**Also declining** interpolated-await scanning and deferred-await dataflow (`var pending = Foo(ct); … await pending;`). Neither shape occurs, and both need real parsing rather than pattern.

On that last point generally: every round here has been a variant of "a regex cannot see X," which is true and unbounded. I've taken the ones that are cheap or genuinely wrong. The principled answer to the rest is the Roslyn analyzer this issue's own step 4 suggests — that would see types, inheritance, and dataflow properly instead of approximating all three. Happy to file it as the follow-up if you want the guard to grow further rather than hold the line at zero.

46 tests now. Reverting all eight source fixes still fails the guard naming exactly those eight; script lane green at 829.

---

rodoHasArrived:
Round four, pushed `890027acc`. Three of six taken — and notably **two of them are regressions from round three's own fixes**, which is the clearest signal yet about where this is heading.

**Fixed**

- **`.Token` over-match.** Round three replaced the name allowlist with harvested declarations but kept a blanket `[\w.]*\.Token`. That treats `await auth.RefreshAsync(session.Token)` as a cancellation path and would report a handler with no cancellation in it at all. `X.Token` now counts only when `X` is a declared `CancellationTokenSource`.
- **Deferred callbacks.** Round three added `ThrowIfCancellationRequested` as a reachability source by searching the whole try body, so `service.Register(() => ct.ThrowIfCancellationRequested())` marked a body reachable even though the callback runs later, outside the catch. Only inline throws count now.
- **`var` aliases.** `var requestToken = context.RequestAborted;` now carries the token onward.

The first two are **false positives** — the direction that actually hurts, since this gates CI and would block valid endpoint code. Neither fires on the current tree; they're latent, and endpoint code changes.

**Declining the rest.** `when (!(ex is not OperationCanceledException))`, `(ex is not OperationCanceledException) == false`, and `when (ex is object)` are pathological spellings with zero occurrences, and reading them correctly means *evaluating* boolean expressions rather than matching them. Treating synchronous token-bearing calls as cancellation sources would broaden reachability to any `Helper(ct)` call — trading the remaining false negatives for a fresh class of false positives, which is precisely the trade the `.Token` expression just had to be walked back from.

---

**I'm holding the line here, and I'd rather say why plainly than keep going.**

Four rounds, each one a variant of "a regex cannot see X." That's true and it's unbounded — but the concrete evidence is that round three's fixes *caused* two of round four's findings. A text scanner chasing C#'s full grammar doesn't converge; each widening creates new surface, and the last two rounds have been net-neutral at best on real risk while adding two false-positive regressions.

What the scanner was built for, it did: it found **five real defects** the manual read missed (`BackfillValidationEndpoints.cs:188/230`, `SymbolEndpoints.cs:248`, `IngestionJobEndpoints.cs:73/117`) and it holds the surface at zero with 51 tests, three of which now pin the *absence* of false positives.

The remaining gaps all need the same three things — types, inheritance, and dataflow. That's a Roslyn analyzer, which is exactly what step 4 of #2618 proposes, and it deserves to be its own change rather than more pattern bolted onto this one. Happy to file it.

Reverting all eight source fixes still fails the guard naming exactly those eight. Script lane green at 834.

---

rodoHasArrived:
Filed the analyzer follow-up as #2705.

It records what the scanner actually earned (the five defects it caught that the manual read missed), the specific gaps grouped by what they need — type identity/inheritance, dataflow, expression evaluation, parsing — and the non-convergence evidence that round three's fixes caused two of round four's findings.

It also carries forward the two judgement calls that were arrived at from real code here and are easy to get wrong on a reimplementation: `when (!ct.IsCancellationRequested)` is a correct idiom rather than a defect, and an earlier `OperationCanceledException` clause exempts the chain regardless of its body — otherwise the SSE heartbeat handlers in `StreamEndpointHelpers.cs` and `WorkstationEndpoints.Stream.cs` get flagged for doing the right thing.

Plus the integration facts: no analyzer project exists yet, `.globalconfig` is an empty scaffold, central package management applies, and the 51 test cases should be ported before the scanner is deleted.

---

Issue #2695, opened 2026-08-15
The API surface is served and documented but unversioned: 1 of ~1,168 routes carries a version


rodoHasArrived:
## Summary

Meridian serves an OpenAPI document and its route surface is discoverable — but essentially nothing is
versioned, so there is no stability promise for anything binding to it.

## What exists

**An OpenAPI document is served.**

```
src/Meridian/UiServer.cs:428-442   AddSwaggerGen(... SwaggerDoc("v1", new OpenApiInfo ...))
src/Meridian/UiServer.cs:566-573   UseSwagger(); UseSwaggerUI(... "/swagger/v1/swagger.json" ...)
docs/reference/api-reference.md:116  advertises /swagger
```

Endpoints carry `.WithName`/`.WithTags`/`.Produces<T>`, so the generated document is reasonably rich.
Discovery — the larger half of the problem — is solved.

There is also a strong *internal* contract: 857 route constants in
`src/Meridian.Contracts/Api/UiApiRoutes.cs`, mirrored into
`src/Meridian.Ui/dashboard/src/lib/ui-api-routes.generated.ts`, keeping the browser and WPF clients in
sync without hand-editing.

## The finding

**One route family carries a version, and it duplicates rather than replaces.** `MapRiskRoutes` maps
seven endpoints:

```
GET  /rules                              GET  /escalations
GET  /rules/{ruleName}/status            POST /escalations/{escalationId}/approve
GET  /rules/{ruleName}/config            POST /escalations/{escalationId}/deny
PUT  /rules/{ruleName}/config
```

and it is invoked **twice** — at `/api/risk` (`RiskEndpoints.cs:78`) and `/api/v1/risk` (`:80`). Both
remain live, so the versioned group is a mirror of an unversioned twin rather than a migration off it.

Every other family is unversioned `/api/<family>`. The Swagger document is titled "v1" while describing
routes whose paths carry no version, so the document's version and the routes' stability are unrelated.

*On counting:* anyone re-deriving the numerator/denominator should enumerate the composed
`EndpointDataSource` rather than grepping — the authorization-sweep test already does this and is a
reasonable place to borrow from. Grep-based counts here mix units (`MapGroup` calls vs routes vs `Map*`
call sites) and are unreliable.

**A second, unused renderer also exists.** `ApiDocumentationService`
(`src/Meridian.Platform/ApiDocumentation/ApiDocumentationService.cs:49-51`) generates its own spec and
[...]

PR #2710, opened 2026-08-15
Compute coverage from an allowlist of reference roots (#2703)

### LINKED ISSUE #2703: Two governed coverage dashboards score prose mentions as documentation
## Summary

`docs/status/api-contract-coverage-dashboard.{md,json}` and `docs/status/coverage-report.md` both score
a symbol as documented when its name appears anywhere in the documentation corpus — including documents
that describe no contract at all. Neither metric can distinguish "documented" from "mentioned."

## How this was found

By tripping it. A draft document was added under `docs/product/` that documented no API and no type —
it discussed several route paths and class names in prose while analysing something else. Regenerating
both dashboards moved both scores:

- **API contract coverage:** `POST /api/auth/accounts/{username}/password-reset` flipped
  `Gap` → `Documented`, and the total went **249 → 250 of 621**. The only new occurrence of that route
  was one prose sentence explaining why an authorization sweep misclassifies it.
- **Public type coverage:** documented types went **2,791 → 2,801**. All ten additions were incidental
  mentions. Four of them — `FileDropRouter`, `FilePermissionsOptions`, `FileSearchResult`,
  `FileToDelete` — appeared **only inside a passage arguing that those classes are not persistent
  stores.** Arguing four classes should not be counted caused them to be counted as documented.

That draft has since been dropped, so the inflation is not in the repository. The behaviour is
reproducible on demand:

```bash
# add any markdown file under docs/ that merely names a route path and a few public type names
git diff docs/status/api-contract-coverage-dashboard.md docs/status/coverage-report.md
```

## Why it matters

Both are governed artifacts, committed to the repo, regenerated by CI, and read as evidence of
coverage. Three consequences:

1. **Gameable by accident.** Any prose naming a symbol raises the score. Nobody has to intend it — in
   both cases above it happened while writing about something else, and in one case while arguing the
   opposite.
2. **One-directional drift.** Documentation volume grows, and paths and type names get named constantly
   in design docs, reviews, ADRs, and changelogs. Scores rise regardless of whether documentation was
   written, so the trend lines are actively misleading.
3. **Unusable for the decision they exist to support.** "250 / 621 endpoints documented" and
   "2,801 / 8,796 types documented" are unknown mixtures of documented and merely-named symbols, so
   neither can prioritize documentation work.

That two independent dashboards share the flaw points at the **corpus rule** as the thing to fix, rather
than either metric individually.

## Suggested direction

1. Restrict both generators' corpora to documentation that describes contracts — `docs/reference/**`
   and equivalents — excluding `docs/product/**`, `docs/plans/**`, reviews, and assessments.
2. Or, more robustly, require the symbol to appear in a structured block (a table row with
   method/path/request/response; an API reference entry) rather than free text.
3. Best, for the API metric: once an OpenAPI document is served for the versioned surface (#2695),
   derive coverage from whether an endpoint has described request/response schemas in the spec — which
   is what "documented" should mean here.

Expect both baselines to drop when this is fixed. That drop is the point; the current numbers are not
coverage measurements.

## Scope note

The generators live in `build/scripts/docs/**`, a governed surface under the phase-scope gate, so a fix
needs a phase-scoped PR rather than a docs-only one.

rodoHasArrived:
## Summary

Both coverage generators now compute from an **allowlist of roots that exist to describe contracts** — `docs/reference` for the API dashboard, plus `docs/generated/database` for the type report — instead of scanning all of `docs/` and subtracting prose.

Fixes #2703.

| Metric | Before | After |
| --- | ---: | ---: |
| Documented endpoints | 251 / 621 | **199 / 621** |
| Documented workstation contracts | 117 / 917 | **34 / 917** |
| Documented items (type report) | 2,821 (32.1%) | **1,227 / 8,783 (14.0%)** |

The drop is the point — the issue's words. The old figures were an unknown mixture of documented and merely-named symbols; the new ones mean something statable: *described in reference documentation*.

## How this converged on the allowlist

This PR went through seven review rounds, and the history is the argument. The first version excluded `docs/product` + `docs/plans` wholesale; review found a blueprint under `docs/plans` whose contract section defines 18 shipped contracts verbatim. The second version excluded only `docs/product`; review found a second contract-bearing blueprint there. The third classified per document by contract headings; review found a delivery plan that enumerates a DTO's complete field set under an ordinary heading — *in order to argue the DTO is missing something*.

Each denylist rule was narrower than the last and each still leaked, because these roots interleave description and argument inside single documents. No cheap syntactic proxy separates "describes a contract" from "argues about one". The allowlist (the issue's suggestion 1) ends the per-file adjudication: naming a symbol in a review, brainstorm, roadmap, or delivery plan no longer moves the score.

Guard-rails carried forward:
- **`docs/generated/database/**` stays in the type corpus** — it is the PostgreSQL data-object catalog with field-level reference documentation; an earlier over-exclusion of it marked 763 genuinely documented types as gaps. Verified present: 67 corpus documents, 34 reference + 33 catalog.
- **The self-referential report exclusions are kept** even though the allowlist makes them unreachable, so widening it later cannot silently reintroduce the feedback loop.
- **The report's remediation now names only scanned roots.** Its old primary advice — run DocFX — wrote to a gitignored path the corpus doesn't (and must not) scan, so following it could never move the number. The advice now says so explicitly. A test pins that no gitignored root may enter the corpus.
- `CLAUDE.md`/`README.md` left the type corpus: project orientation prose, not reference.

## Testing performed

Python, verified locally end to end. **Both** test locations for these generators pass — `tests/scripts/test_api_contract_coverage_dashboard.py` (55 tests) and `build/scripts/docs/tests/test_generate_api_contract_coverage_dashboard.py` (6 tests); a mid-review commit broke the second one precisely because it is wired to no CI lane, which is recorded below. Script lane green at 842; regeneration idempotent on a second pass; ratchets exit 0.

The corpus rule is pinned as an exact tuple in both generators, the retained roots have counterweight tests, and behavioural tests assert a route in `docs/reference` counts while the same route in `docs/product` does not.

## Findings recorded along the way (not fixed here)

- **#2713** — `docs/generated/documentation-coverage.md` is a tracked generated artifact no automation regenerates; it disagreed with `docs/status/coverage-report.md` on `main` before this PR.
- **8 of the 10 suites under `build/scripts/docs/tests/` are wired to no CI lane** (the two wired ones are invoked by name, not discovery). Same shape as audit finding P9. Closing it is a 4-line `scripts/ci.sh` change — governance, so it needs explicit approval.
- **`test_validate_docs_structure` fails on Linux** (asserts a Windows path separator), pre-existing on `main`.

## Scope note

Declared `phase:PR6` (this PR touches `build/**`, `tests/**`, and `docs/**`, all within PR6's cumulative scope). Verified locally: `enforce_phase_scope.py --phase PR6` passes against this diff; PR1 correctly rejects it.

## Safety review

## Governance changes

- [ ] `.github/workflows/**`
- [ ] `.github/CODEOWNERS`
- [ ] `AGENTS.md`
- [ ] `scripts/ci.sh`


rodoHasArrived:
Both findings checked. The second was right and I've fixed it in `cacf3880d`; the first is real but predates this PR.

## Taken: `docs/plans` was an over-exclusion

Verified per file rather than per root, which is what I should have done the first time:

| Root | File | Endpoints | Contracts | Kind |
| --- | --- | ---: | ---: | --- |
| `docs/plans` | `security-master-passport-workbench.md` | 0 | **18** | contract section |
| `docs/product` | seven prose files | 3 | 13 | reviews, brainstorms, todo list, delivery plan |
| `docs/product` | `portfolio-cash-ladder-blueprint-2026-07.md` | 0 | 1 | contract section |

**All 18 of `docs/plans`' contract credits come from that one blueprint** — excluding the root dropped 18 legitimate credits and zero prose ones. That is precisely the 763-type mistake `generate-coverage.py` already carries a comment about, and I repeated it at smaller scale *in the change that quotes it*. `docs/plans` is back in both corpora.

Verified the specific contract you named is credited again — `SecurityMasterRevisionStateDto` now reads `documented: true` in the regenerated dashboard.

`docs/product` stays excluded on the evidence above: seven of its eight crediting files are prose, supplying every endpoint credit and 13 of 14 contract credits. The known cost is the single blueprint credit. I've taken that deliberately — under-counting by one beats over-counting by thirteen — and written it into the constant's comment so it is a *known* false negative rather than a silent one. I'd rather not add a per-file classifier for one credit; the heading conventions across these roots are too varied (`## Interface & API Contracts`, `## 7. Contracts summary`, `### REST API surface`) for a marker rule to be anything but another silent-failure mode.

Corrected totals against the pre-change baseline:

| | Before | Was (over-excluded) | **Now** |
| --- | ---: | ---: | ---: |
| Endpoints | 251 | 247 | **247** |
| Workstation contracts | 117 | 84 | **103** |
| Documented types | 2,821 | 2,630 | **2,671** |

## Not taken here: `documentation-coverage.md`

The inconsistency is real, but it is **pre-existing and not caused by this PR**. On `main`, before any of my changes:

```
docs/generated/documentation-coverage.md   2735 / 8765  (31.2%)
docs/status/coverage-report.md             2823 / 8786  (32.1%)
```

The *totals* differ — 8,765 against 8,786 — so the two were generated at different commits, not merely with different corpora.

The cause is that `run-docs-automation.py` registers `generate-coverage` with `args: ["--output", "docs/status/coverage-report.md"]`, which overrides the CLI's documented default. So nothing in the automation pipeline ever regenerates the tracked default artifact, and it has been drifting independently.

Regenerating it here would fold ~90 items of unrelated pre-existing drift into a PR about corpus rules, and would not fix the reason it goes stale. The actual defect is a tracked generated artifact with no generator wired to it — worth its own change, either pointing the automation entry at both outputs or deleting the orphan. Happy to file that; say the word and I'll do it rather than leave it implied.

Script lane green at 838, docs regeneration idempotent on a second pass.

---

rodoHasArrived:
Right again, and this one exposes a flaw in my own evidence rather than just my conclusion. Fixed in `e90d08f70`.

`web-ui-report-run-stream-blueprint-2026-07.md` does carry a `## 7. Contracts summary` documenting `ReportingRunAuditTrailDto` and `NullReportingRunNotifier`. I had classified it as prose — because my classifier searched for the literal string `"Interface & API Contracts"`. A single-marker rule. Which is the exact fragility I cited, one comment earlier, as my reason for *not* building a classifier.

So my "seven of eight product files are prose" figure was an artifact of a bad detector, and I used it to justify keeping the root exclusion. The real cost was 4 contract credits, not 1.

**Testing the objection instead of asserting it.** Widening the marker to a heading pattern separates the files cleanly:

| | Contract headings |
| --- | ---: |
| Six genuinely-prose files (reviews, brainstorms, todo list, delivery plan) | **0** |
| `portfolio-cash-ladder-blueprint-2026-07.md` | 5 |
| `web-ui-report-run-stream-blueprint-2026-07.md` | 1 |

The discriminator I dismissed as unworkable works on the corpus it actually has to work on. That was checkable in about a minute and I should have checked before arguing.

**A root no longer excludes anything on its own.** Membership marks a root for *checking*; a document is dropped only when it carries no contract heading. Both generators do this — `generate-coverage.py` gains a `PROSE_DOC_PREFIXES` list separate from its hard exclusions, since the self-referential reports stay unconditional.

Verified both contracts you named now read `documented: true`, alongside `SecurityMasterRevisionStateDto` from the previous round.

| | Baseline | Wholesale roots | **Now** |
| --- | ---: | ---: | ---: |
| Endpoints | 251 | 247 | **247** |
| Workstation contracts | 117 | 84 | **107** |
| Documented types | 2,821 | 2,630 | **2,703** |

Every credit still removed is prose-derived, which is what the issue asked for.

Three tests added: the heading pattern accepts `Contracts summary` and `Interface & API Contracts` but rejects `API notes`, and a blueprint written into a prose root keeps its route credit. Script lane green at 840; regeneration idempotent.

The remaining known weakness is honest and stated in the code: a blueprint that describes contracts under some heading this pattern doesn't match will still be dropped. That fails in a direction I can name, unlike the root rule, which had no way to tell the two kinds apart at all.

---

rodoHasArrived:
Correct on the facts, and I'm going to stop patching and step back, because four rounds have found four counterexamples and that is the finding.

## On this specific case

`w9-close-out-delivery-plan-2026-08.md:708-713` does enumerate `StatementRunCreateDto`'s complete field set. But it's worth reading what the passage is doing — it sits under `### W9-INGEST-009` in a delivery plan, and the enumeration exists to argue a defect:

> **But "already carried" holds only for the governed intake flow — one live caller cannot supply a scope at all.** … its full field set is the broker, source institution, … **no fund profile, ledger book, accounting period, or as-of date anywhere in it.**

That is the same shape as the example the issue opens with — "four of them appeared only inside a passage arguing that those classes are not persistent stores." A complete field enumeration written to prove something is *missing*. Whether that counts as documentation is genuinely arguable, and I don't think either answer is obviously right.

## The pattern across four rounds

| Round | Rule | Counterexample |
| --- | --- | --- |
| 1 | Exclude `docs/product` + `docs/plans` | `security-master-passport-workbench.md` — 18 real credits |
| 2 | Exclude `docs/product` only | `web-ui-report-run-stream-blueprint` — `Contracts summary` |
| 3 | Heading allowlist | `w9-close-out-delivery-plan` — ordinary heading |
| 4 | ? | — |

Each fix was narrower than the last and each still leaked. The reason is that these roots interleave the two kinds *within* documents, and no cheap syntactic proxy — root, filename, heading, or field-set enumeration — separates "describes a contract" from "argues about one." I no longer believe a denylist of any granularity converges here, and continuing to widen it is how a metric ends up with rules nobody can state.

## What the alternative measures out at

The issue's suggestion 1 in its strong form — restrict the corpus to documentation that describes contracts, rather than subtracting prose from everything:

| Corpus | Endpoints | Contracts |
| --- | ---: | ---: |
| Everything (today's `main`) | 251 | 117 |
| **`docs/reference` only** | **199** | **34** |
| `docs/reference` + `generated/database` | 199 | 41 |
| `docs/reference` + `architecture` | 216 | 52 |
| This PR as it stands | 247 | 107 |

The allowlist numbers are smaller, but they mean something statable: *"described in reference documentation."* No per-file judgment, no counterexamples to find, and the rule fits in a sentence. The denylist number is larger and means "named somewhere we haven't excluded yet," which is what the issue objected to in the first place.

## Where I've left it

The branch currently holds the round-3 state (247 / 107). I'm not pushing a round-4 patch, because which way to go is a decision about what the metric should assert, not a defect I should keep guessing at — and I've already changed its meaning three times on your behalf today.

Three ways forward, in my order of preference:

1. **Switch to the allowlist** — I'll rewrite this PR to compute from `docs/reference` (± `generated/database`) and let the number drop to ~199/34. Crisp definition, no ongoing adjudication.
2. **Land it as-is** — strictly better than `main` on prose inflation, with false negatives I can't bound.
3. **Close it** and let #2695's OpenAPI surface carry the metric, per the issue's own suggestion 3, which is the only version where "documented" has a mechanical definition.

Happy to do any of them. I'd rather ask than iterate a fifth time.

---

rodoHasArrived:
Fixed in `b27400e84`. I broke that test and didn't notice, and the reason I didn't notice is worth more than the fix.

## The fix

The test's intent survives the change — it's the counterweight to the generated-report case directly above it, asserting that a *hand-written* document does count. It now writes into `docs/reference` and is renamed `test_reference_docs_count_as_documentation`. I added a companion, `test_a_non_reference_root_does_not_count_as_documentation`, which keeps the original `docs/development` case and asserts the opposite — the new rule stated from the other side, so the change is pinned in both directions rather than just relaxed.

## Why CI didn't catch it

`build/scripts/docs/tests/` is a second test location for this generator, and the script lane I've been reporting green all day (`run-script-tests.py`) discovers only `tests/scripts/`. So my "840 tests green" was true and irrelevant to the suite I was breaking.

Checking the whole directory, **eight of its ten suites are wired to no CI lane at all**:

| Suite | Wired? |
| --- | --- |
| `test_check_codex_memory` | ✅ `documentation.yml` |
| `test_pilot_readiness_dashboard` | ✅ `golden-path-validation.yml` |
| `test_generate_api_contract_coverage_dashboard` | ❌ |
| `test_check_ai_handoff`, `test_check_ai_inventory`, `test_generate_structure_docs`, `test_markdown_generation_lint`, `test_run_docs_automation_pilot_optin`, `test_scan_todos`, `test_validate_docs_structure` | ❌ |

The two that are wired are invoked individually by name, not by discovery, so adding a suite to that directory wires it to nothing.

This is the same gap audit finding P9 recorded for `tests/scripts/` — "~65 of 75 script-test suites were wired to no CI lane and several had rotted unnoticed" — which `run-script-tests.py` exists to close. That runner already takes a `--start-dir`, so pointing a second invocation at `build/scripts/docs/tests` would close it here too. I haven't done that in this PR because it's a `scripts/ci.sh` edit, which is governance and needs your explicit approval; say the word and it's a four-line change.

## One pre-existing failure, not mine

While checking, `test_validate_docs_structure` fails on Linux — it asserts a Windows path separator:

```
AssertionError: 'docs\ai\guide.md' not found in "... docs/ai/guide.md: Missing front-matter field ..."
```

Confirmed pre-existing by stashing this branch's changes and re-running against `main`. It's presumably been failing on every non-Windows machine since it was written, which nothing noticed because the suite isn't wired. I've left it alone — it's unrelated to this PR and wants its own fix.

All ten suites in the directory now pass except that one. Script lane still green at 840.

---

rodoHasArrived:
Taken — this is the best catch of the six. Fixed in `f3d181ae0`.

The allowlist made the report's own advice unfollowable. Its primary remediation for the public-type gap is *"generate API docs with DocFX (`docfx docfx.json`)"*, and `docfx.json:29` writes to `docs/docfx/api`, which I wasn't scanning. An operator could run the recommended command and watch the number not move. **A report that recommends an action which cannot change the figure printed beside it is worse than one that prints the wrong figure** — the first wastes someone's afternoon, the second just misinforms.

`docs/docfx/api` is generated API reference built from XML doc comments, so it is reference documentation by exactly the reasoning that keeps `docs/generated/database` in the corpus. Added to both generators.

I also dropped *"or relevant architecture docs"* from the smaller-gap branch, since `docs/architecture` isn't scanned and naming it would send someone to write in a place that can't count.

## Checked the blast radius rather than assuming

Your comment made me verify the allowlist hadn't quietly broken the other categories too. It hadn't — each is matched against its own target file rather than the general corpus:

| Category | main | This branch |
| --- | --- | --- |
| Public Classes / Interfaces | 2726 / 8317 | **1133 / 8314** |
| API Endpoints | 85 / 329 | 85 / 329 |
| Configuration Options | 1 / 129 | 1 / 129 |
| Provider Implementations | 0 / 0 | 0 / 0 |
| ADR Implementations | 11 / 11 | 11 / 11 |

Only the type metric moved, which is what was intended — but I hadn't confirmed it before you raised this, and if `_check_config_documentation` had used the shared corpus I'd have silently zeroed a second metric.

Added a test pinning that the remediation targets sit inside the corpus, so a future recommendation can't point somewhere unscanned without failing.

Both dashboard test locations pass, script lane green at 841, regeneration idempotent.

---

Six rounds on this PR, and the shape of them is worth recording for whoever reads it later: five found real defects in changes I'd verified and described as safe, and the two most valuable — this one and the second test location — were about *consequences* of a correct-looking change rather than the change itself. My verification kept confirming the thing I'd built and missing what it touched.

---

rodoHasArrived:
Right, and it means my previous commit was wrong rather than incomplete. Reverted and replaced in `6ea189ee0`.

**It didn't work.** The loader enumerates `DOC_FILE_EXTENSIONS = (".md",)` and DocFX emits per-type metadata as `.yml`, so my allowlist entry admitted only the tracked `index.md`. I changed the directory, didn't change the extension, and then wrote that the remediation was repaired. One commit after saying my verification "kept confirming the thing I built and missing what it touched," I did it again.

**It also shouldn't work**, which is the part that matters more. `docs/docfx/README.md` states the generated metadata is *"rebuilt by DocFX and ignored by Git"*, and `.gitignore:235` confirms it:

```
docs/docfx/api/*.yml
```

Scanning a gitignored root would make coverage depend on whether whoever ran the generator had run DocFX first — CI and a developer would compute different numbers from the same commit. That is exactly the reproducibility trap that already cost this branch a `regenerate-docs` failure over `todo-scan-results.json`. Loading the `.yml` extension, as the comment offers as one option, would have walked straight back into it.

And it would have re-inflated the metric. DocFX enumerates the public surface of every project under `src/`, so its metadata names types whether or not anyone documented them. Crediting that is the inflation #2703 was filed about, arriving through a gitignored side door.

**So the advice is fixed rather than the corpus.** The public-type recommendation now reads:

> Add reference entries under `docs/reference/` for the types that carry contracts. Running DocFX does not move this metric: it emits browsable output from XML doc comments into `docs/docfx/api/*.yml`, which is gitignored and so is not part of the scanned corpus.

That is honest about what running DocFX achieves — browsable API output, worth having, not this measurement.

Two tests: no gitignored root may enter the corpus, and the public-type recommendation must name a scanned one. The first would have caught my last commit.

Types unchanged at 1130 / 8314; script lane green at 842; both dashboard test locations pass; regeneration idempotent.

---

Issue #2702, opened 2026-08-15
Duplicate-helper ratchet is name-based, so a renamed clone walks straight past it


rodoHasArrived:
Found while working #2614's remaining helpers (PR #2700).

## Summary

`build/scripts/ci/check-duplicate-helpers.py` detects duplication by **name** — it tracks `NormalizeOptional`, `RequireText`, `FirstNonBlank`, `TrimOrNull`, `FirstNonEmpty`. A copy that is spelled differently is invisible to it, even when the body is character-for-character the canonical helper.

That is not hypothetical. Grouping every `Normalize`-named declaration in `src/` by alpha-equivalent body turns up a cluster of **9 declarations** whose body is exactly `TextPrimitives.NormalizeOptional`:

```csharp
string.IsNullOrWhiteSpace(value) ? null : value.Trim()
```

The ratchet reported "baseline satisfied" the entire time those existed.

## Why it matters

The ratchet's own docstring states the risk it exists to manage:

> The cost of that is not the duplicated lines, it is that a copied helper is free to drift. `NormalizeOptional` had already split into three behaviours under one name and signature […] Nothing failed at the time it diverged, which is what this check is for.

A rename is the cheapest possible way to produce exactly that outcome, and it is also the *natural* thing to do when copy-pasting a helper into a class that wants a shorter name. So the check is blind precisely along the axis its rationale is about.

## Why "just track `Normalize` too" is the wrong fix

`Normalize` has **55 declarations across 46 files, and 41 distinct bodies**. It is a generic verb covering genuinely different domain functions, not 55 copies of one thing. Adding it to `TRACKED_HELPERS` would fire on 41 legitimate functions to catch 9 real clones — the check would immediately be worth ignoring.

The same holds for the other names in that measurement:

| Helper | Declarations | Distinct bodies |
| --- | ---: | ---: |
| `Normalize` | 55 | 41 |
| `NormalizeCurrency` | 19 | 9 |
| `FormatBytes` | 15 | 8 |

## Suggested direction

Detect by **body**, not name: canonicalise each `private`/`internal` helper body (strip comments, fold literals, rename identifiers positionally) and flag any that is alpha-equivalent to a helper `TextPrimitives` already owns, whatever it is called. That catches renamed clones and cannot fire on a function that merely shares a name.

A prototype of the canonicaliser is in PR #2700's working notes; the same alpha-equivalence approach was used on #2612 to get an honest divergence count, so the technique is already proven in this repo.

[...]

Issue #2705, opened 2026-08-15
Replace the endpoint-cancellation text scanner with a Roslyn analyzer


rodoHasArrived:
Follow-up to #2618, step 4 ("Consider an analyzer or review-checklist item so new endpoints default to the helper"), and to the four review rounds on PR #2699.

## Summary

PR #2699 added `build/scripts/ci/check-endpoint-cancellation.py`, a regex/brace scanner that reports endpoint handlers whose generic `catch` can swallow a client disconnect. It works, and it earned its keep — but it has reached the limit of what pattern matching can do, and the remaining gaps all need the same three capabilities: **types, inheritance, and dataflow**. That is a `DiagnosticAnalyzer`.

## What the scanner achieved (so it isn't rebuilt from zero)

Worth recording, because the rule itself is validated even if the implementation should change:

- The manual read of #2618's 62 `catch (Exception` sites found **3** real defects.
- The scanner, once widened to bare `catch { }`, found **5 more** the eye had missed — `BackfillValidationEndpoints.cs:188/230`, `SymbolEndpoints.cs:248`, `IngestionJobEndpoints.cs:73/117`. Three of those answer **200 with silently incomplete data**; two answer **400 "Invalid request body"** for a caller who merely hung up.
- It now holds the endpoint surface at zero, with 51 tests.

So the rule is worth enforcing. The question is only how.

## Why pattern matching can't finish the job

Four review rounds on #2699 each surfaced a new "the regex cannot see X". The decisive evidence is that **round three's fixes directly caused two of round four's findings**:

| Round 3 change | Round 4 consequence |
| --- | --- |
| Replaced the token allowlist with harvested declarations, kept a blanket `[\w.]*\.Token` | `await auth.RefreshAsync(session.Token)` classified as a cancellation path — a false positive on code with no cancellation in it |
| Added `ThrowIfCancellationRequested` as a reachability source via whole-body search | `service.Register(() => ct.ThrowIfCancellationRequested())` marked the body reachable, though the callback runs later outside the catch |

Both were fixed, but that is the shape of a non-converging loop: each widening creates new surface. False positives are the harmful direction here, because this gates CI and would block valid endpoint code.

The gaps still open, grouped by what they actually require:

**Type identity and inheritance** — a semantic model resolves these; a regex approximates them by name:
- `catch (SystemException)` and `catch (Exception)` both catch cancellation, but so would any other base type someone picks.
[...]

PR #2762, opened 2026-08-16
Detect renamed duplicate helpers by body, not just name

rodoHasArrived:
## Summary

Partial fix for #2702 (detection; migration is follow-up tranche work). The consolidated-helper ratchet tracked five names, so a copy pasted under any other name was invisible — and renaming is the natural move when pasting a helper into a class that wants a shorter word. `check-duplicate-helpers.py` now also compares by **body**: each `private`/`internal` declaration is canonicalised — comments stripped, parameters and `var`/`foreach` locals renamed positionally, string/char/numeric literals folded, single-`return` blocks unwrapped so a braced copy equals its expression-bodied twin — and flagged when it is alpha-equivalent to a body `TextPrimitives` owns, whatever the copy is called. Member and type names are deliberately left intact, so `IsNullOrEmpty` variants and behaviour-changing copies (the historical case-folding split) stay distinct. Owner bodies are read from `TextPrimitives.cs` at runtime, so the check tracks the shared surface without a second hand-maintained list.

The scan found **35 renamed clones across 33 files** — not the 2 the issue projected, which had measured only `Normalize`-named declarations. Across all names the clones include `CleanOptional`, `NullIfBlank`, `NullIfWhiteSpace`, `NormalizeId`, `NormalizeKey`, `NormalizeText`, `NormalizeNullable` (all exact `NormalizeOptional` bodies), one `RequireTrimmed` = `RequireText` (differing only in the folded message literal), and — pointedly — a `NormalizeOptionalToken` whose body is exactly `NormalizeOptional`, the near-name shape the name-based check deliberately exempts on the assumption the rename "says what it does". All 35 are baselined under a new `body_clones` map for migration tranches; tracked names remain the name ratchet's jurisdiction so one copy is never double-counted.

## Reason

Issue #2702: the ratchet's own rationale is that a copied helper is free to drift, and a rename is the cheapest way to produce exactly that outcome invisibly. Adding generic names to the tracked list instead would not work — `Normalize` alone has 41 distinct bodies across 55 declarations, so a name rule would drown 9 real clones in 41 false fires (and, as the body scan shows, would still miss the other 26 under names the issue never measured).

## Testing performed

- 9 new tests (18 total in the suite): renamed clone detected; braced single-return equals expression form; `IsNullOrEmpty` variant and behaviour-changing (case-folding) copy are **not** clones; guard clone matches despite a different exception message; loop clone matches with renamed locals; tracked names stay the name ratchet's jurisdiction; the canonical home is not a clone of itself; and the committed `body_clones` baseline must describe the real tree — the same decorative-baseline guard the name map already has.
- Spot-verified three detections by reading source: `ExecutionOperatorControlService.NormalizeOptionalToken`, `FileEvidenceArtifactStore.RequireTrimmed`, and the double-declared `NormalizeOptionalTenantTemplateId` (two classes in one file — the count of 2 is real).
- Live run: `Consolidated-helper ratchet: 76 private declaration(s) across 67 file(s), 35 baselined renamed body clone(s) (baseline satisfied)`, exit 0; name-based counts unchanged.
- Full script-test lane: 879 tests, 0 failures.

## Safety review

- Pure detection change: the name-based counts and baseline entries are byte-identical; nothing that passes today starts failing, because every found clone is baselined.
- The scanner skips anything it cannot bracket cleanly (a miss is a quieter failure than a false fire on mangled input), and `--update-baseline` regenerates both maps together.

## Governance changes

None. No workflow, CODEOWNERS, or CI-policy surface is touched.

---


rodoHasArrived:
Round-three disposition: four of the five findings are fixed in the latest push (dollar-width raw interpolation, BCL type-alias normalization, redundant-paren stripping, and the signature gate + segment-jumping masker — the standalone check drops from ~43s to ~5.6s).

The fifth — keying on `[CallerArgumentExpression]` and default values — is deliberately not taken. Including them in the clone key would un-flag `FileEvidenceArtifactStore.RequireTrimmed`, which is a verified genuine clone: same guard, same trim, message-only difference, and every call site passes the parameter name explicitly. The flag's contract is "a human migrates this to the shared helper", and replacing an attribute-less guard with `RequireText` strictly improves `ParamName` reporting; reconciling call shapes is that migration's job, not the detector's. The scanner's docstring now states the scope boundary this implies: the canonicaliser targets accidental copies (paste, rename, reformat, respell), not deliberate evasion — semantic equivalence beyond that is a Roslyn-analyzer problem, the same escalation boundary drawn for the endpoint-cancellation guard in #2705.

---

rodoHasArrived:
Round-five disposition: the return-type finding is taken — return types now join the clone key under the same canonical spelling rules as parameter types, and all 36 baselined clones survive the stricter key (they all return the owner's type).

The bare-declaration finding was implemented and measured before declining. With the keyword requirement made optional (line-anchored, statement-keyword lookahead as the replacement guard), the header matched control flow — the owner map came back with "methods" named `foreach` and `if` — and the standalone check quadrupled from ~5s to ~21s. The keyword requirement is what keeps statement text out of a regex-based scanner; a lookahead list is just an incomplete enumeration of C#'s statement forms. Bare no-modifier instance helpers are also structurally implausible clone carriers: a helper worth sharing is static by nature, and an instance copy is already a different function. The header comment records this boundary so the experiment isn't silently re-run.

---

rodoHasArrived:
Round-seven disposition: four of six taken in the latest push — `global::` qualifiers fold with the CLR aliases; a verbatim `@`-string is never a raw string, so `@""""` no longer consumes the rest of the file (that one was a genuine parsing bug that could hide every later clone in a file); an identifier after `.` is a member and is protected from the parameter rename; and the identifier classes now follow Unicode word rules.

Two declined, both at the documented scope boundary:

- **Custom types named `String`**: distinguishing a project-defined `String` from `System.String` requires symbol resolution, which a text scanner does not have — this is exactly the Roslyn-checker boundary stated in the module docstring. The failure mode also requires a repo to define its own `String` *and* give a helper over it the owner's exact body; if that ever happens, the human reviewing the flag will see it immediately.
- **Instance/override methods with owner bodies**: these are correctly flagged. A contract-required override whose body duplicates `NormalizeOptional` migrates by *delegation* — `public override string? Clean(string? v) => NormalizeOptional(v);` — which satisfies both the base contract and the ratchet. Restricting matching to static declarations would hide real copies of the shared logic.

At seven rounds the detector has reached the precision a text scanner supports: further findings constructible only by deliberate single-case authorship (as opposed to accidental copy patterns or false fires on code this repo would plausibly contain) fall under the documented scope and will be declined on that basis.

---

PR #2797, opened 2026-08-19
Reserve gross exposure so concurrent orders cannot overshoot the ceiling (3/5)

rodoHasArrived:
## Summary

PR 3 of 5 in the "repair irreversible correctness boundaries" series, following #2526 and #2794.

`GrossExposureRule` measured every order against a snapshot of *settled* positions and reserved nothing. Nothing serialises pre-trade validation — `CompositeRiskValidator` holds no lock, and the order manager validates each submission independently — so orders evaluated beside one another each observed the same headroom below the ceiling and all passed, breaching a `Critical` portfolio-wide limit none of them breached alone. Routed orders cannot be recalled, so that breach is only unwound at whatever the market charges.

- **The check and the consumption are now atomic.** The rule implements `IReservingRiskRule`, re-measuring inside the ledger lock against the exposure other in-flight orders hold and taking the order's own in the same step. The mechanism was already present and already documented for exactly this hazard — `IReservingRiskRule`'s own remarks state that splitting the two "would let concurrent submissions each observe room below the cap and all commit, overshooting it." The exposure rules had simply never adopted it, leaving `OrderRateThrottle` as the only reserving rule.
- **Reservations are qualified by fund account.** The portfolio-wide gross ceiling counts every account's in-flight notional, because gross is a property of the whole book. Direction-aware projection consults the order's *own* account only: netting across accounts here would repeat the mistake `SymbolExposure.ResolveSignedExposureFor` already refuses to make on filled positions, letting a flat fund sell "against" another fund's long. Orders with no fund account share one unattributed pool — counted in the portfolio total, never in an attributed account's signed exposure.
- **One ledger is shared across the validator.** The rule defaults to a private ledger so existing composition keeps working, which means the workstation registration supplies the shared instance explicitly; a per-rule ledger would make each order the only order it can see. Registered as a singleton, matching the validator it serves.

`EvaluateAsync` stays side-effect free as its contract requires — the composite validator evaluates every rule before deciding, so a rule that consumed capacity there would charge orders a later rule then blocked.

## Reason

The series invariant is at-least-once correctness with fail-closed behaviour at irreversible boundaries. A breached exposure ceiling is irreversible in the strongest sense in this series so far: the orders are live at a venue and cannot be recalled. This was a time-of-check/time-of-use gap that widens with submission rate — the more concurrent the flow, the more orders can spend the same headroom.

## Scope, and what is deliberately not covered

- **The routed-to-filled window is unchanged.** A reservation is held from evaluation until the caller settles it, and both outcomes release: a rolled-back order never reached a venue, and a committed order's exposure becomes the portfolio's to account for. Holding past commit would need a release driven by order-terminal state, and `SettleRiskReservations` runs once at routing with no such seam — without one the ceiling would tighten permanently with every order, which is a worse failure than the gap being closed. What closes here is concurrent evaluations spending one order's headroom, which is the part that scales with submission rate.
- **Per-symbol ceilings do not reserve yet.** `PositionLimitRule` and `SymbolConcentrationRule` are cross-order capacity too, but keyed per symbol and — for position limits — denominated in quantity rather than notional, so they need a separate ledger dimension rather than this one. Called out in the runbook rather than left implicit.
- **`OrderNotionalRule` is correctly excluded.** It gates each order independently against a per-order cap and shares no capacity across orders, so there is nothing to reserve.

## Testing performed

- [x] `bash scripts/ci.sh` lanes runnable in this environment completed successfully
- [x] Relevant integration tests were added or updated
- [ ] GitHub Actions `quality-gate` passed — running on this head

Notes:

- Branch restarted from `main` at `3b883a3f9` per the merged-PR rule; PR 2 is present in `main` via that merge, and no commits are stacked on merged history.
- `ExposureReservationTests` adds 7 tests: concurrent orders cannot each spend the same headroom; settlement returns headroom exactly once and is idempotent across repeated/mixed settlement; in-flight exposure counts across accounts for the portfolio ceiling; exposure is attributed to the owning account and not the unattributed pool; a refused order takes no capacity; an unconfigured ceiling reserves nothing; and `EvaluateAsync` remains side-effect free.
- Mutation-verified: with in-flight exposure dropped from the projection, both concurrency tests approve the second order and the pair breaches the ceiling.
- `Meridian.Tests.Risk` and `Meridian.Tests.Execution` pass 840/840 locally; `dotnet format whitespace --verify-no-changes` and the file-size ratchet both clean.
- Scope note: as with PR 2, the series brief named this area without enumerating defects, so the specific work was derived from current source against the series invariant. The supporting evidence is in the code — `IReservingRiskRule` documents the hazard, and only the rate throttle had adopted it.

## Safety review

- [x] No unrelated changes were included — the diff is confined to the risk rule, its new ledger, the composition that shares it, tests, and the docs describing it

## Governance changes

Check every governance file modified:

- [ ] `.github/workflows/**`
- [ ] `.github/CODEOWNERS`
- [ ] `AGENTS.md`
- [ ] `scripts/ci.sh`

https://claude.ai/code/session_01MzeZrUtwwf9CnTdrSzALti

---
_Generated by [Claude Code](https://claude.ai/code/session_01MzeZrUtwwf9CnTdrSzALti)_


rodoHasArrived:
All three findings are correct, and the first one is fatal to the premise rather than to the implementation. I've reverted both commits in `f5d8aae12`; the branch now matches `main` and this PR should be closed rather than fixed.

**The defect this PR claimed to fix does not exist.**

Pre-trade validation is already serialised. `OrderManagementSystem` holds `_preTradeReservationGate` across validation and registration, and says so directly: *"Validation and the registration that reserves this order's exposure happen under one gate so a concurrent order cannot slip through against the same pre-order snapshot."* I checked `CompositeRiskValidator` for a lock, found none, and concluded validation was unserialised — without checking the caller.

The snapshot is also not settled positions only, which was my other load-bearing assumption. `AggregatePortfolioExposureProvider.ApplyWorkingOrderExposure` folds accepted-but-unfilled orders into it, aggregated per `(symbol, account)` specifically so a long with two working sells cannot double-reserve — handling the netting case more carefully than my ledger did. Its summary describes the exact scenario I set out to prevent: *"a burst of working orders cannot each observe a flat book and collectively breach a ceiling that none of them breaches alone."*

**Keeping it would have been worse than doing nothing.** As the P1 on double-counting sets out, a working order counted once in `snapshot.GrossExposure` and again in `totalGrossInFlight` projects a book no fill subset can reach, so a `Critical` rule would reject valid orders and trip the execution circuit breaker — halting the desk on arithmetic rather than risk. The $40k + $30k → $110k example is exactly right.

The other two findings were real bugs in the reverted code and I'm not disputing them: the account-wide signed total applied to a single symbol would have let an in-flight AAPL buy offset an MSFT sell, and reserving the pre-lock notional rather than the value the locked remeasurement approved would have under-reserved on a price move. Fixing them would only have refined a mechanism that shouldn't exist.

Worth recording for the series: the existing design here is stronger than I credited, and the guarantee is split across two layers — the gate in the OMS and the working-order fold in the exposure provider — with neither visible from the rule itself. That's what I missed by reading `GetSnapshot()` as "settled positions."

---

